PR #9, opened 2017-08-18
Make more struct the conditions for return address validation.

unkadoug:


src/tool/hpcrun/unwind/x86-family/x86-validate-retn-addr.c
```diff
@@ -414,7 +416,7 @@ validation_status
 validate_return_addr(void *addr, void *generic)
 {
     unwindr_info_t unwr_info;
-	if( !uw_recipe_map_lookup(addr, &unwr_info) ) {
+	if (!uw_recipe_map_lookup(addr, &unwr_info) ) {
```
jmellorcrummey: I think that the test in validate_return_addr should not just check for false as a return from the map lookup. It should be like the test in deep_validate_return_address: 

if (!uw_recipe_map_lookup(addr, &unwr_info)  || unwr_info.treestat == NEVER)

jmellorcrummey:
reviewed. seems fine.

PR #42, opened 2017-12-06
Change the unwind step process for libunwind to match the x86 and ppc…

unkadoug:
…64 behaviors.  Currently,

libunwind finds a recipe and then applies it in a step.  To match the others, it must first
apply a (previously found) recipe and then find and saving the next recipe.  This change
addresses a problem related to the computation of normalized ips.

Change the mixing of x86 native unwinding and libunwind-based unwinding to adapt to the
libunwind changes.

I'd appreciate testing help.


unkadoug:
Passed hpcstruct test on ppc, x86, arm.  Verified by John to fix normalized ips issue.

PR #54, opened 2018-02-13
Mult kernel

laksono:
merging fixes for perf into master.


jmellorcrummey:
the time has come to merge mult-kernel into master. while there are some minor issues (e.g. partial unwinds on ARM are not properly reported), mult-kernel is much better than master and deserves to be merged.

PR #61, opened 2018-02-27
Have only one bitree_uwi_t * in a ilmstat_btuwi_pair_t, rather than o…

unkadoug:
…ne for

each unwinder.  Having one for each unwinder means that freeing a bitree_uwi_t
that is part of a skip list node for the skip list for one unwinder tries
to do something regarding the bitree_uwi_t for the other unwinder, which
can't be right.


unkadoug:
I believe this fixes something that is wrong.  I am less sure that it fixes the bug seen on ray.llnl.org.

I can walk the skip list search to get to the node with invalid values and a null pointer that we dereference, and in trying to find a way for a skip list node to have those values, I have only thought of the possibility of something being freed unexpectedly, and then I noticed that our free function in this case could conceivably be trying to free two ilmstat_btuwi_pairs, even though I expect that one of them is null and so this doesn't really explain anything.

PR #92, opened 2018-08-02
Fixing problems with failed building intervals

laksono:
just in case building intervals crashes within uw_recipe_map, we wrap the code with siglong jmp.


mwkrentel:
I looked at this wrt sigsetjmp/siglongjmp.

First, let me check.  If the code crashes in the inner scope, it does
make sense to fix the unwind state and continue, right?  That is, you
don't have to go all the way to the top and drop the sample, right?

I can't find anything exactly wrong with the code paths.  It can be
tricky to know what is the right buffer (if any) to jump to in all
cases.  For example, you have to be careful not to get inside the
inner scope, long jump to the outer scope, then take another segv and
think that the inner buffer is still valid.  But I can't find a case
where this is handled incorrectly.

Parts of the code are kinda gross and not how I'd do things, but they
were gross before Laksono got there.  I could take a pass and try to
clean up a few things.  But it wouldn't really be much cleaner, and I
can't point to anything that would actually run differently.

So ok, don't let me hold you up from moving forward.

PR #104, opened 2018-08-29
Hpcprof mpi bcast

jmellorcrummey:
overhaul communication in hpcprof-mpi
- replace hand-crafted broadcast tree with MPI_bcast
- avoid unnecessary barriers


jmellorcrummey:
merged hpcproftt man page and hpcproftt/args.cpp manually

PR #138, opened 2019-05-25
Many metrics: control hide/show metrics from hpcrun

laksono:
**Goal**: To control which metrics to show/hide from hpcrun.

**Problem**: hpcprof initialize all metrics from hpcrun as "visible" and set all helper metrics as "invisible". Later on, hpcprof will only output "visible" metrics to the database. Hence, all "invisible" metrics from hpcrun will always be visible in the database. 

**Solution**: In hpcprof/hpcprof-mpi, add another flag "isTemporary" to specify which metrics are from hpcrun, and which metrics are temporary in hpcprof. the later then output metrics that are not temporary.


laksono:
@jmellorcrummey, @Jokeren : I submit a pull request to control hide/show metrics from hpcrun. It works in my case, but I don't know if it works or not in GPU branch. Can you please review and test it ? This shouldn't remove place holder nodes, but you never know. Thanks.

laksono:
It works with default options on hpcprof and hpcprof-mpi but fails when using `-M threads` flag:
```
$ mpirun -np 2 ~/pkgs/laksono/hpctoolkit/bin/hpcprof-mpi -M stats -S amg2006.hpcstruct  hpctoolkit-amg2006-measurements
...
ERROR: [std::bad_alloc] std::bad_alloc
```
Needs further investigation.

PR #141, opened 2019-06-15
Add the config option '--with-mbedtls=path' to test for libmbedcrypto.a

jmellorcrummey:
to provide md5 and SHA functions from mbedtls.

We require libmbedcrypto.a to be compiled with -fPIC to be used inside
a shared library.


jmellorcrummey:
@mwkrentel I want to merge this into master so that building master with -mbedtls does not report an error. This is one step short of requiring mbedtls, which happens in the master-gnu branch.

mwkrentel:
It's not an error, only a warning and harmless.
Plus, you won't even see it when building with spack.

mwkrentel:
Gpu now merged into master which includes this.

PR #151, opened 2019-11-26
Fix unassociated memcpy cct nodes

Jokeren:
1. It is safer to have a separate routine to handle memcpy2.

2. In some cases, if we leave the system to figure out a dtoh or a htod copy, the api callback itself does not specify the copy direction. Therefore by default we assign a <gpu copy> placeholder to these apis, such as `cuMemcpy`.


jmellorcrummey:
I accepted the change, but I think it would be better if we put in the placeholders at the end rather than anticipating what placeholders might be needed. Adding lots of unused placeholders (as we do for OMPT and I think the runtime API) is inefficient.

PR #153, opened 2019-11-28
Master gpu trace unified incomplete merge

Jokeren:
Previously, event synchronization activities are not implemented. 

Besides, on Power9 machines, gpu activities on the same stream may have overlapped intervals. So we adjust the start timestamp of an activity as max(last_activity_end + 1, current_activity_start).

There are other features have been implemented in master-gpu-trace branch but not merged. I will only add essential fix, such as the timestamp adjustment but not merge all of them.


jmellorcrummey:
thanks for adding the new event functionality

PR #157, opened 2019-12-12
Vdso cleanup

mxz297:
This PR contains:

1. Merge current "master" into "gpu".

2. Remove vdso hash value from the viewer.

image

@jmellorcrummey Instead of normalizing the file name that contains the vdso hash value at file name construction time, I decided to normalize the vdso hash value when generating the XML file. The code for constructing the file name is quite complex. 

Meanwhile, we normalize the load module name when generating XML file. So, I think it should be also fine to normalize the file name when generating XML file.


jmellorcrummey:
@mxz297 normalizing the file name when producing the XML is a fine choice.

PR #161, opened 2019-12-19
Exporting metric description into the database

laksono:
The current hpcrun is already equipped with metric description, but it doesn't gather description from the sample-source. Hence, the content of the description is not used in hpcprof.
This patch includes:
- hpcrun: gathering metric description from perf_events sample-source.
- hpcrun: writing the description into hpcrun file
- hpcprof: exporting it to experiment.xml

Caveat: due to XML limitation, a metric description cannot include special charaters (need to convert to escape sequence).


jmellorcrummey:
tested. harmless to old hpcviewer. needs new hpcviewer to reveal metric descriptions. 

hpcviewer pane and pull down menu item names should be changed. instead of "metric properties", it should say "metric descriptions".

PR #168, opened 2020-01-18
Handle duplicated event id

Jokeren:
The `eventId` field is a little confusing.

https://docs.nvidia.com/cupti/Cupti/annotated.html#structCUpti__ActivityCudaEvent

> uint32_t CUpti_ActivityCudaEvent::eventId
> A unique event ID to identify the event record.
> 

EventId is unique with regard to `cuEventCreate` but not 'cuEventRecord'. That is, when an event is created (cuEventCreate), it is assigned a unique id. When an event is pushed into a stream (cuEventRecord), it triggers the creation of a `CUpti__ActivityCudaEvent` record. If the event is pushed again, CUPTI API creates another `CUpti__ActivityCudaEvent` record with the same event id but a different correlationId.

### DIFF
--- a/src/tool/hpcrun/gpu/gpu-event-id-map.c
+++ b/src/tool/hpcrun/gpu/gpu-event-id-map.c
@@ -171,10 +171,22 @@ gpu_event_id_map_insert
  uint32_t stream_id
 )
 {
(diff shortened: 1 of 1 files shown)


jmellorcrummey:
looks good!

PR #175, opened 2020-01-22
Add count metrics

Jokeren:
In the tutorial slides, we use average number of bytes copied to determine if pined memory optimization could be applied.


src/tool/hpcrun/gpu/gpu-metrics.c
```diff
@@ -220,6 +221,22 @@ gpu_metrics_attribute_metric_time_interval
...
+gpu_metrics_attribute_metric_count
+(
+ cct_node_t *cct_node,
+ int count_index,
+ int count
+)
+{
+  metric_data_list_t *metrics = hpcrun_reify_metric_set(cct_node, count_index);
+
+  // convert from ns to s
```
jmellorcrummey: this comment should be deleted

src/tool/hpcrun/gpu/gpu-metrics.c
```diff
@@ -307,6 +325,9 @@ gpu_metrics_attribute_mem_op
 
   gpu_metrics_attribute_metric_time_interval(cct_node, time_metric_index, 
 					     (gpu_interval_t *) m);
+  
+  // number of kernel launches
```
jmellorcrummey: this comment should be 

// increment the count of this kind of operation

jmellorcrummey:
the code looks good. only the comments need to be updated.

src/tool/hpcrun/gpu/gpu-metrics.c
```diff
@@ -583,6 +606,16 @@ gpu_metrics_default_enable
...
+#undef CURRENT_METRIC 
+#define CURRENT_METRIC GCOUNT
+
+  INITIALIZE_METRIC_KIND();
+
+  FORALL_GCOUNT(INITIALIZE_SCALAR_METRIC_INT)
+
+  FINALIZE_METRIC_KIND();
+
 // Memcpy metrics
```
Jokeren: The following metrics are for cudaMemoryAlloc/Free APIs but not memcpy APIs.

Jokeren: And we could remove GMEM from the default metrics for now because actually we do not monitor memory alloc/dealloc events. CUPTI does not provide a correlationId field in CUpti_ActivityMemory.

src/tool/hpcrun/gpu/gpu-metrics.c
```diff
@@ -583,6 +606,16 @@ gpu_metrics_default_enable
 
   FINALIZE_METRIC_KIND();
 
+// Execution count metrics
+#undef CURRENT_METRIC 
+#define CURRENT_METRIC GCOUNT
+
+  INITIALIZE_METRIC_KIND();
+
+  FORALL_GCOUNT(INITIALIZE_SCALAR_METRIC_INT)
```
Jokeren: There's a design issue.

All metrics, including GMEM, GMSET, GXCOPY, and GICOPY are `INDEXED_METRIC`, while to increase GMEM_COUNT, GMSET_COUNT, GXCOPY_COUNT, and GICOPY_COUNT, we need `SCALAR_METRIC`. 

Thus, I do not recommend to add COUNT as an index in existing GPU API metrics. Rather, we shall keep the current solution that uses a single GCOUNT metric, or init GMEM_COUNT, GMSET_COUNT, GXCOPY_COUNT, and GICOPY_COUNT separately as different metrics.

src/tool/hpcrun/gpu/gpu-metrics.c
Jokeren: Not sure if roctracer API manifests correlation ids for memory alloc/free events.

src/tool/hpcrun/gpu/gpu-metrics.c
jmellorcrummey: Adding @mxz297 to the conversation:

Regarding GMEM: 

Without correlation ID, we can't properly attribute them if we use the activity API. Can't we attribute them using the callback API? (I think this would be a change from the implementation now, but I haven't checked)

Regarding using COUNT with INDEXED_METRICs:

If we use type indices 0..N, why is it a bad idea to record the count using index N+1? Alternatively, we can use type indices as 1..N+1 and record the count as index 0. 
Why do I propose this? I think keeping all of the relevant things together is useful. If we monitor one of the types, we will monitor all of them (count included).

I don't know whether roctracer has correlation ids for allocation and deallocatiion.

Jokeren: > Without correlation ID, we can't properly attribute them if we use the activity API. Can't we attribute them using the callback API? (I think this would be a change from the implementation now, but I haven't checked)

That needs a lot of implementation changes. Inside `cupti-api.c`, we have to manually identify every memory allocation API, read parameters, and calculate the time spent in these APIs. I suggest to add it in the next version.

> If we use type indices 0..N, why is it a bad idea to record the count using index N+1? Alternatively, we can use type indices as 1..N+1 and record the count as index 0.

Yes, we could. My thought was that the INDEXED_METRICs are used for activity **types**, reserving the last slot for COUNTs might be confusing for future development. If we decide to do so, we would better add some comments.

Jokeren: Another problem of adding COUNT as the last index.

> FORALL_GSYNC(INITIALIZE_INDEXED_METRIC_REAL)

GSYNC metrics are REAL but not INT, adding 1 to a REAL metric does not have effect, though theoretically we could add 1.0 to GSYNC_COUNT.

jmellorcrummey: It would be OK to add 1.0 to counts as a real.

src/tool/hpcrun/gpu/gpu-metrics.c
jmellorcrummey: We can certainly remove GMEM from default metrics since we don't monitor them now.

Jokeren:
@jmellorcrummey , Since @laksono has merged the formula branch, please check if this branch is ready to merge.

Issue #189, opened 2020-01-30
BLOCKTIME event crashes


laksono:
Tested platforms: x86 and Arm
Branch: master
Cause: Incorrect use of metric kind. The block-time  and Linux perf_events  have different metric-kind. When genevent_set is triggered, it queries the number of metrics, and it returns 0 since block-time event is not part of Linux perf_event metric kind.

Call stack from the core file:
```
(gdb) bt
#0  0x0000ffff8235cc48 in record_sample (current=0xffff7ec4da90, mmap_data=0xffffc384c3f0, context=0xffffc384c770, sv=0xffffc384c5b0) at ../../../../src/tool/hpcrun/sample-sources/perf/linux_perf.c:549
#1  0x0000ffff8235e100 in perf_event_handler (sig=38, siginfo=0xffffc384c6f0, context=0xffffc384c770) at ../../../../src/tool/hpcrun/sample-sources/perf/linux_perf.c:1163
#2  0x0000ffff822deab8 in monitor_signal_handler (sig=38, info=0xffffc384c6f0, context=0xffffc384c770) at signal.c:218
#3  <signal handler called>

(gdb) l
544
545       // check if this event is multiplexed. we need to notify the user that a multiplexed
546       //  event is not accurate at all.
547       // Note: perf event can report the scale to be close to 1 (like 1.02 or 0.99).
548       //       we need to use a range of value to see if it's multiplexed or not
549       info_aux->is_multiplexed    |= (scale_f>PERF_MULTIPLEX_RANGE);
550
551       // case of multiplexed or frequency-based sampling, we need to store the mean and
552       // the standard deviation of the sampling period
553       info_aux->num_samples++;
(gdb) p info_aux
$1 = (metric_aux_info_t *) 0x0
```
`info_aux` is null because it doesn't get the correct number of metrics. Hence, it allocates zero byte:
```
 930   int num_metrics = hpcrun_get_num_metrics(lnux_kind);
 931
...
 941
 942   size_t mem_metrics_size = num_metrics * sizeof(metric_aux_info_t);
 943   metric_aux_info_t* aux_info = (metric_aux_info_t*) hpcrun_malloc(mem_metrics_size);
 944   memset(aux_info, 0, mem_metrics_size);
 945
 946   thread_data_t* td = hpcrun_get_thread_data();
 947
 948   td->core_profile_trace_data.perf_event_info = aux_info;
```

mxz297:
[...]

Issue #144, opened 2019-08-22
Plans for merging master-gpu branch to master


mxz297:
@jmellorcrummey @mwkrentel @Jokeren @laiwei-rice 

In the next a few weeks, we hope to merge the master-gpu branch to the master branch. This merge will be a beast. Since I am new to the HPCToolkit code base, I will need a place for tracking the problems I have and explanations I got from corresponding people. That's reason for creating this issue. 

This merge will bring 30K lines of addition and 3K lines of deletion to the master branch. The exact files that will be changed can be found in the following link:

https://gist.github.com/mxz297/4fd1982d1a247e985973ce009f24504a

The first step of my plan is for everyone who may have committed code to the mater-gpu branch to take a look at the above link and then write a summary about his code changes. Note that the merge has over 500 commits and github is only willing to show the first 250 commits, so I cannot easily identify who has changed what. So, I will need each person to identify his changes and write descriptions.

The descriptions I am hoping for should include the following items:
* What's the problems are you solving? 
* What are your steps for solving these problems in a conceptual level?
* What are the source files changed corresponding to each conceptual step?
* What are the existing code you changed for solving your problems? Describe why existing code does not work for you and use the master branch the definition for existing code.

I will then carefully read your descriptions and read the corresponding source code. My goal is to have a decent understanding of the code, identify inconsistent code design, and identify hard-to-understand code as soon as possible. 

It may seem time consuming to type descriptions in a github issue rather than explaining things in person. But to deal with this large amount of code, I absolutely need to track the descriptions and reference them multiple times. So, please write descriptions in detail and I will carefully read them and cross check with the code. 

A concurrent step is to test the merge in a black-box way. This merge should be able to profile normal cpu code (results should be similar to master branch), NVIDIA GPU code (similar to master-gpu), and improved support for openmp (similar to master-gpu). And a typical test workflow is to manually see whether the hpcviewer/hpctraceviewer are showing similar results before and after the merge. 

I think we need to first determine several testing benchmark for profiling for the three major capabilities (normal cpu code, NVIDIA GPU code, and OpenMP support). I can then determine a strategy for splitting the task. 

Finally, I will start with making sure the merge will compile on various platforms.

Jokeren:
Maybe I will merge this branch into master-gpu first. It just contains a few additional features for instruction mix analysis and only has few changes to the existing master-gpu code.

https://github.com/HPCToolkit/hpctoolkit/tree/master-gpu-instruction-analyzer

[...]

PR #194, opened 2020-02-03
hpcrun: fix issue #189 (BLOCKTIME event crashes)

laksono:
Quick fix for issue #189 where profiling with -e BLOCKTIME causes segmentation fault in some cases, and memory corruption (but doesn't crash) in other cases.

Current solution: All linux_perf's custom-based events should be based on perf_event metric kind. Otherwise linux perf_event manager doesn't know how many metrics (or events) are there.
Long term solution: Linux perf_event custom-based events should be ideally more independent (ideally it should be a plugin).

### DIFF
--- a/src/tool/hpcrun/sample-sources/perf/event_custom.h
+++ b/src/tool/hpcrun/sample-sources/perf/event_custom.h
@@ -54,7 +54,7 @@ struct event_thread_s;
 struct perf_mmap_data_s;
 
 // callback functions
-typedef void (*register_event_t)(struct event_info_s *);
+typedef void (*register_event_t)(kind_info_t *kb_kind, struct event_info_s *);
 typedef void (*event_handler_t)(struct event_thread_s*, sample_val_t , struct perf_mmap_data_s* );
 
 
--- a/src/tool/hpcrun/sample-sources/perf/kernel_blocking.c
+++ b/src/tool/hpcrun/sample-sources/perf/kernel_blocking.c
@@ -210,12 +210,8 @@ kernel_block_handler( event_thread_t *current_event, sample_val_t sv,
  * - context switch metric to store the number of context switches
  ****************************************************************/
 static void
-register_blocking(event_info_t *event_desc)
+register_blocking(kind_info_t *kb_kind, event_info_t *event_desc)
 {
-  static kind_info_t *kb_kind;
-
-  kb_kind = hpcrun_metrics_new_kind();
-
   // ------------------------------------------
   // create metric to compute blocking time
   // ------------------------------------------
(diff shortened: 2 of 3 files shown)


mwkrentel:
There's something related to this that I seriously don't understand.

In #144 I pointed out that there appears to be a type mismatch for the
function `hpcrun_set_new_metric_info_and_period`.

`hpcrun/metrics.h` declares this as follows with a `kind_info_t *`
first arg.

```
int hpcrun_set_new_metric_info_and_period(kind_info_t *kind, const char* name,
                                          MetricFlags_ValFmt_t valFmt, size_t period, 
					  metric_desc_properties_t prop);
```

But there are files: `cuda.c`, `generic.c`, `papi.c`, `tst.c` that use
it without the kind pointer.

```
sample-sources/cuda.c:

      hpcrun_set_new_metric_info_and_period(strdup(buffer),
                                            MetricFlags_ValFmt_Int,
                                            self->evl.events[i].thresh);
```

What am I missing here?
How does this even compile?

I asked @jokeren to fix this back in October, but I don't see that
ever happened.

@jmellorcrummey   @Jokeren   @laksono

Jokeren:
@mwkrentel , these mismatches should be fixed. I will touch base with @laksono today.

And I will also discuss with @jmellorcrummey for the Fatbin.cpp issue.

Sorry for missing the request.

mxz297:
@mwkrentel I have seen similar problems before where undefined functions do not trigger any compilation or linking error. I believe we are using certain linker flags to suppress undefined functions errors.  I find a post online that describes a few linker flags to achieve this goal:

https://stackoverflow.com/questions/5555632/can-gcc-not-complain-about-undefined-references

Our configure file as shown below contains lots of appearance of a `allow_undefined_flag` flag. I guess this is the flag that controls whether we ask the linker to ignore undefined functions. 

https://github.com/HPCToolkit/hpctoolkit/blob/master/configure#L26312

I don't quite know the benefits of suppressing linking errors for undefined functions. I don't think we need such suppressions for calling functions in dynamically loaded libraries.

mxz297:
@jmellorcrummey @laksono I tested this PR and read its code changes. This PR looks good.

Jokeren:
@mxz297, There's another PR needs test.

https://github.com/Jokeren/hpctoolkit/tree/fix_metric_kind

mxz297:
@Jokeren Is there a PR or/and issue for it? I would like some contexts to understand what's the purpose of the branch.

Jokeren:
There are other problems just like this BLOCKTIME issue. As mentioned by Mark, `hpcrun_set_new_metric_info_and_period` is not correctly used.

Do I just push my commits to this PR, or create another PR?

mxz297:
I would prefer new PR to document the details.

Jokeren:
The new PR is here:

https://github.com/HPCToolkit/hpctoolkit/pull/201

mxz297:
@Jokeren Thanks!

PR #196, opened 2020-02-04
Fix CUPTI hang problem when fork is used

Jokeren:


jmellorcrummey:
this pull request was incorporated into #297

PR #198, opened 2020-02-05
Add a configure-time requirement on Boost system + graph.

blue42u:
This one burned me while trying to minimize my Boost footprint. Without this, `./configure` will just happily let some (very long) link commands fail mysteriously.

### DIFF
--- a/configure
+++ b/configure
@@ -20305,7 +20305,8 @@ BOOST_COPY=no
 # dyninst would use (but no more).  But copy-libs won't fail on a
 # missing library, so it doesn't have to be exact.
 
-BOOST_COPY_LIST='atomic chrono date_time filesystem system thread timer graph regex'
+BOOST_COPY_LIST='atomic chrono date_time filesystem graph regex system thread timer'
+BOOST_NEEDED_LIST='graph regex system'
 
 { $as_echo "$as_me:${as_lineno-$LINENO}: checking for boost" >&5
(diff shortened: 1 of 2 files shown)


mwkrentel:
First, what's your reluctance to run autoreconf?  You should not
commit changes to configure.ac without running autoreconf.
And it needs to be the correct rev of autotools.  I can show you what
to do.

How exactly did you "minimize" your boost footprint?
What was the link line that failed?

Let me think about it a little more.  But I think you're missing
'regex' from the needed list.  And I want to check that there aren't
other libraries that we use.  (That was the reason for linking
anything that Dyninst uses.)

blue42u:
Just an annoyance of the build system. We can discuss at another time.

Minimizing the set of Boost libraries that I need to build, since I do that myself. It took me an hour or two scanning through the `hpcstruct-bin` link line before I realized that `-lboost_graph` was missing, thus the change to have `./configure` sanity check it.

Boost 'graph' automatically links to 'regex', so 'regex' doesn't need to be present on the link line. Unless we ever change to static link `hpcstruct-bin`.

Similarly, since we dynamic link Dyninst, the only bits we need to include on the link lines are whatever is pulled by Dyninst's headers. In Boost, the only one should be 'system' for OS-specific error handlers, everything else should be behind a Dyninst (real) function.

blue42u:
Boost regex added to the link line as promised.

PR #199, opened 2020-02-05
Add partial Valgrind support (configure option and Struct annotations).

blue42u:
Adds `--with-valgrind` and `--enable-valgrind-annotations`. Compatible with Spack's `valgrind` package.


src/tool/hpcrun/memory/hpcrun-malloc.h
```diff
@@ -73,7 +74,7 @@
 extern "C" {
 #endif
 
-#if ! defined(VALGRIND) || defined(_IN_MEM_C)
+#if 1 // ENABLE_VG_ANNOTATIONS != 1 || defined(_IN_MEM_C)
```
jmellorcrummey: this looks useless and perhaps should be omitted.

blue42u: By "this" do you mean the entire #if...#else block, or just this change? Enabling the "Valgrind-safe" allocation methods causes segfaults at runtime.

jmellorcrummey: The purpose of hpcrun_malloc is to support allocation in signal handlers. The purpose of hpcrun_malloc_safe is to support allocation outside signal handlers using our allocator in a way that won't be interrupted by signals.

The #else of the #if block simply rewrites these as malloc. That won't work in hpcrun. Is there any reason why the #if 1, the #else and the block that follows, and the #endif should be retained. If not, they should be discarded.

blue42u: Doesn't seem to be a reason. Removed.

blue42u:
@jmellorcrummey @mwkrentel Are there any remaining issues in this PR?

jmellorcrummey:
I have no concerns with the PR. Unless Mark voices concerns by the end of the day, I will accept the PR.

mwkrentel:
One small thing, @blue42u could you add a 1-line AC_MSG_NOTICE message
near the bottom of configure.ac to display the path for valgrind, or
else no or empty (like the others).  That way, the summary will show
if you're using valgrind or not.

They're supposed to be alphabetical, but the cuda, cupti and papi-c
are misplaced.

mwkrentel:
Actually, one other thing, does --enable-valgrind-annotations add
anything extra that --with-valgrind doesn't have?

Are there any tradeoffs with the extra annotations?

My question is what value the second knob adds, or should there
just be one knob.

blue42u:
Notice message added.

The separation is to allow disabled Valgrind annotations when using `--with-spack`. The theory is each annotation spends 6 unoptimized arithmetic cycles, and there are usually 1-2 annotations per synchronization operation per thread. They also act as full fences, so they prevent a number of optimizations.

In practice, I don't know exactly how much of a performance effect they have, I've just followed the conservative approach. Since new-Prof has a whole bunch of 'em stitched in it would be nice to have the separation.

mwkrentel:
But you could use the --with-valgrind value to set the AC_DEFINE
symbol ENABLE_VG_ANNOTATIONS.  So, if --with-valgrind is set (or
implicitly taken from --with-spack), then annotations are turned on,
and if not set, then the symbol is not defined.

That's what I meant by "added value."  There are no degrees of
valgrind, it's either on or off, there is no extra on.

For better or worse, that's how we currently treat --with-cuda.  If
you use --with-spack and cuda is in the spack install dir, then it
uses it, unless overriden with --without-cuda.

Really, this is just a way to reset the default to not use valgrind
unless you specifically ask for it, which I think is a good thing.

Another way to enforce the same default (no, unless specifically asked
for), would be to remove the implicit "look for this in --with-spack
and use it if it's there."  That way, even if valgrind was in the
spack dir, it wouldn't use it unless you added --with-valgrind=PATH.

I guess I would slightly lean that way, but I don't have strong
feelings about it.  You can leave it this way, if you like.

Either way, I approve.

Btw, the AM_CONDITIONAL variable OPT_ENABLE_VG_ANNOTATIONS
is for conditional Makefile settings.  I don't see where you ever use this.
But maybe I missed one.

mwkrentel:
Btw, I don't see how to review.  I was trying to give you a green check,
but all I accomplished is that I deleted and re-added myself as a reviewer.
:-(

PR #201, opened 2020-02-07
Fix incorrect use of hpcrun_set_new_metric_info_and_period and missing hpcrun_close_kind

Jokeren:
`hpcrun/metrics.h` declares `kind_info_t *` as the first arg.

```
int hpcrun_set_new_metric_info_and_period(kind_info_t *kind, const char* name,
                                          MetricFlags_ValFmt_t valFmt, size_t period, 
					  metric_desc_properties_t prop);
```

But there are files such as  `cuda.c`, `generic.c`, `papi.c`, and `tst.c` that use it without the kind pointer.

Besides, `hpcrun_close_kind` is not called in some places after `hpcrun_metrics_new_kind` is used.

### DIFF
--- a/src/tool/hpcrun/sample-sources/cuda.c
+++ b/src/tool/hpcrun/sample-sources/cuda.c
@@ -329,16 +329,19 @@ METHOD_FN(process_event_list, int lush_metrics)
 
   hpcrun_pre_allocate_metrics(nevents + num_lush_metrics);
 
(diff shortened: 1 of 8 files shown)


mwkrentel:
It compiles and I think all the cases of `hpcrun_set_new_metric_info_and_period()` 
have a `kind_info_t *` parameter.

PR #202, opened 2020-02-07
Fnbounds -- fix bug relating to static binaries

martyitz:


mwkrentel:
Works for me on the one example I tried.

PR #204, opened 2020-02-08
Use clone() rather than fork() for spawning the hpcfnbounds process.

blue42u:
Fixes a number of issues with `pthread_atfork` handlers. Probably only works on Linux.

### DIFF
--- a/src/tool/hpcrun/fnbounds/fnbounds_client.c
+++ b/src/tool/hpcrun/fnbounds/fnbounds_client.c
@@ -124,6 +124,7 @@ char	*outfile;
 #include <err.h>
 #include <errno.h>
 #include <fcntl.h>
+#include <sched.h>
 #include <signal.h>
 #include <stdbool.h>
 #include <stdint.h>
@@ -147,6 +148,8 @@ char	*outfile;
 
 // Limit on memory use at which we restart the server in Meg.
 #define SERVER_MEM_LIMIT  140
+// Size to allocate for the stack of the server setup function, in KiB.
+#define SERVER_STACK_SIZE 1024
 #define MIN_NUM_QUERIES   12
 
 #define SUCCESS   0
@@ -160,6 +163,7 @@ enum {
 
 static int client_status = SYSERV_INACTIVE;
 static char *server;
+static char *server_stack;
 
 static int fdout = -1;
 static int fdin = -1;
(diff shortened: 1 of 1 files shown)


jmellorcrummey:
@mwkrentel You should look at this PR to use clone rather than fork. I am not worried about the Linux specificity.

mwkrentel:
I think the use of `char stack[...];` as a local variable is
technically broken.  Depending on the scheduling, if launch_server()
returns and starts running more code, it will be writing on top of the
new process's stack.

You'll get away with it because there's 1Meg of space before you get
to something that matters.  But still ....

Also, a local variable will add 2Meg to the current stack.  Process
stacks are normally 8Meg, but thread stacks are often smaller.  So,
2Meg is a bit of a risk of blowing stack.

I guess, mmap() a single 1-2Meg region and keep it around for future
launches that need to restart the server.  Don't create a new region
each time.

It's either that or you could use `CLONE_VFORK`.  That would fix the
overwrite problem, but not the exceed stack problem.

Also, stack sizes are normally 8Meg.  It's odd that 'man clone'
doesn't show a way to tell the system what the size of the stack is.
I don't know what they expect or how they tell the difference between
a 16K stack or 8Meg.  Some other arch's have an extra arg for this,
but not ones that we use.

And see if there is some system way of asking do stacks grow up or
down on this system.  It seems gross to create a stack that's twice
the size.

But fix the local variable problem and then I'll say yes.

Btw, what does 'Fixes a number of issues with pthread_atfork handlers'
mean?  It fixes 1 issue.  What are the other N - 1?

blue42u:
I can't really test the stack growth direction: https://stackoverflow.com/a/3572947

As for (some of) your local variable concerns, the child lives in a copy-on-write memory space separate from the parent, so there are no race conditions:
```
              If  CLONE_VM  is  not  set, the child process runs in a separate
              copy of the memory space of the calling process at the  time  of
              the  clone call.  Memory writes or file mappings/unmappings per‐
              formed by one of the processes do not affect the other, as  with
              fork(2).
```

For the blowing stack concern, I think the effective waste is 1MiB (aren't stack offsets hard-coded in the assembly?), but I'm not super comfortable with it either. Its not possible to `mmap` a global array for it though because of the races:
```
       Note that the glibc clone() wrapper function makes some changes in  the
       memory  pointed  to by stack (changes required to set the stack up cor‐
       rectly for the child) before invoking the clone() system call.  So,  in
       cases  where clone() is used to recursively create children, do not use
       the buffer employed for the parent's stack as the stack of the child.
```

I'd like to use the syscall directly to let it handle the stack, but the headache depicted in NOTES of `man clone` is a problem. And `clone3` was just added in the `5.2.x` series, so we can't use that.

Thus, where we sit now.

----
It fixes roughly any issue related to `pthread_atfork` handlers, since now we never `fork`. I wasn't particularly pedantic with my English when writing the OP, really just needed a description for a fairly short patch.

mwkrentel:
> As for (some of) your local variable concerns, the child lives in a
> copy-on-write memory space separate from the parent, so there are no
> race conditions:

D'uh.  Sorry for my blind spot.  I guess I was blinded by the name
'clone.'

There are three main places to allocate memory.  (1) on the heap with
malloc, (2) on the stack, or (3) with anonymous mmap.

(1) is no good, we don't malloc() inside hpcrun.  The issue with (2)
is size.  Long ago, we encountered a thread library that allocated
small, 64K thread stacks.  We had to intercept pthread attributes to
bump that up to 2Meg because PAPI was using 500K.  Here, 'stack' is 4X
the PAPI size, and that's for every thread.

What's wrong with anon mmap (or hpcrun_malloc which uses anon mmap)
a single 1-2 Meg region for this?  We don't share it from parent to
child.  It's only reused if we launch fnbounds and then need to
relaunch it (either fnbounds crashes or runs up a large memory
footprint).

blue42u:
D'uh, forgive *my* blind spot. How does this version look?

PR #208, opened 2020-02-13
Fix issue 68 (incorrect struct ID for single aliens)

laksono:
**Symptom**: hpcprof generates the same flat ID (or struct ID) for the same inlined macro (single alien) from different contexts, which confuses the viewer when creating the flat view.

**Root cause**: Before writing experiment.xml, hpcprof removes redundant procedures by ensuring the tuple `<Load_module, Filename, Procedure, line_number>` is unique. However, for inline macros, we shouldn't remove redundancy. Or perhaps we should add `<Load_module, Filename, Procedure, line_number, "parent_id">` to be unique.

**Solution**: add "`parent_id`" in the tuple to make sure single alien is unique based on its context.

### DIFF
--- a/src/lib/prof/CallPath-Profile.cpp
+++ b/src/lib/prof/CallPath-Profile.cpp
@@ -669,6 +669,18 @@ writeXML_help(std::ostream& os, const char* entry_nm,
         completProcName.append(file_key);
         completProcName.append(":");
 
+        if ((strct->type() == Prof::Struct::ANode::TyAlien) &&
+            strct->name().compare("<inline>")==0) 
+            {
+          Prof::Struct::ANode *parent = strct->parent();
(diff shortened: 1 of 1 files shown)


laksono:
@Jokeren : Can you check with QuickSilver code if this solution works?

jmellorcrummey:
Quicksilver works in part. The code fix for aliens prevents the unholy blending of performance information from different contexts in the bottom up view. However, it somehow interferes with the GPU CCT reconstruction. 

@Jokeren @laksono and I need to discuss this. Below is what I see from the top-down view now. Notice that once we get onto the GPU, we see a flat profile:

image

jmellorcrummey:
Keren and I reviewed it together and we will merge with master and close this branch.

Issue #211, opened 2020-02-14
we missed a place where vdso names need to be prettified


jmellorcrummey:
we need to prettify the vdso load module name when creating unknown procedure names in hpcprof

<img width="636" alt="image" src="https://user-images.githubusercontent.com/3926112/74537229-4e916580-4eff-11ea-9ad1-3d3f1032f7f0.png">

mxz297:
@jmellorcrummey I am having difficulty of reproducing this problem. When I ran my small CUDA example with `hpcrun -t -e REALTIME -e gpu=nvidia` with the latest master branch, the flat view I saw is as follows:

![image](https://user-images.githubusercontent.com/1683713/74610016-17fd4b80-50b5-11ea-98fe-2e2883cfdcfa.png)

You can see that there are not `<unknown procedure>` entries in my viewer. Did you run hpcstruct on the vdso file?

jmellorcrummey:
@blue42u Can you provide Xiaozhu your deep wave measurements directory related to the issue I reported here? Xiaozhu doesn't need any binaries because the attribution problem surfaced with the vdso file in your measurements directory.

mxz297:
@jmellorcrummey 
![image](https://user-images.githubusercontent.com/1683713/74700545-8a5f5000-51c9-11ea-8102-442222174b88.png)

I got the vdso name normalized. However, I find that there is a "unknown load module" in my viewer. I am not sure if it is normal, or if it is caused because I got a moved directory, or if it is a bug.

jmellorcrummey:
I think the unknown load module occurs because there are load modules that are missing. The missing load modules can be found on Jonathon’s home system, but not where you were doing this work. Your result looks good. i will look at the commit carefully to see where changes were needed.

blue42u:
The `<unknown load module>` occurs on my system too, Prof uses the path of the load module if it can't find it (and spits out `<unknown procedure> 0xXXX [LM]` tags).

Those are the bad unwinds, the unwinder lands in a region with apparently no load module. The logs show some very interesting addresses, some of which are obvious errors (like `0x1000`), but some look important (like `0x7f65a28b46fa`, from `libstdc++` at `0x7f65d679af75`).

Which is why I asked at the last meeting exactly how gracefully Run fails if Fnbounds crashes while analyzing a module.

mxz297:
@jmellorcrummey Can we close this issue?

mxz297:
Fixed by #228

Issue #213, opened 2020-02-14
can't unwind out of std::condition_variable::wait on x86


jmellorcrummey:
<img width="660" alt="image" src="https://user-images.githubusercontent.com/3926112/74539171-33285980-4f03-11ea-8d9d-f9c0881af4bc.png">

jmellorcrummey:
@blue42u Please upload a tar file onto box that contains a copy of the following libraries that you used when profiling deep wave (on your home rig?):

```
/lib/x86_64-linux-gnu/libpthread-2.29.so
/usr/lib/x86_64-linux-gnu/libstdc++.so.6.0.28
```

I need to look at the binaries to understand the root cause of the unwind problem you encountered in std::condition_variable::wait.

blue42u:
Here you are: https://rice.box.com/s/fs97ix2c4qabgjvh1wpeldmdxodb6v43

jmellorcrummey:
Do you still have a log file from the deep wave execution in your measurements directory? That would also be helpful for tracking down this problem.

blue42u:
I believe the database you're looking at is the same one I gave Xiaozhu: https://rice.box.com/s/ugwziimczi95o1bv03u2bs3xtki0qh5i.

Here's a log file for from fresh run too: [python3-000000-000-007f0101-498772-0.log](https://github.com/HPCToolkit/hpctoolkit/files/4233711/python3-000000-000-007f0101-498772-0.log).

blue42u:
[This repository has moved to GitLab, please continue this issue there.](https://gitlab.com/hpctoolkit/hpctoolkit/-/issues/213)

PR #215, opened 2020-02-18
Procedure name normalization for VDSO and unknown procedures

mxz297:
Fixes for #211

### DIFF
--- a/src/lib/analysis/CallPath.cpp
+++ b/src/lib/analysis/CallPath.cpp
@@ -127,17 +127,6 @@ std::ostream* Analysis::CallPath::dbgOs = NULL; // for parallel debugging
 static void
 coalesceStmts(Prof::Struct::Tree& structure);
 
-static bool
-isVDSOLoadModule(string name) {  
-  // vdso load module name is in the following format:
-  // vdso/<md5 hash>.vdso
-  size_t ret = name.rfind("/");  
-  if (ret == string::npos) return false;
-  if (name.substr(0, ret) != "vdso") return false; 
-  return name.find(".[vdso]") == name.size() - 7;
-}
-
-
 namespace Analysis {
 
 namespace CallPath {
@@ -528,23 +517,8 @@ overlayStaticStructureMain(Prof::CallPath::Profile& prof,
   } else {
     try {
       lm = new BinUtil::LM();
-      if (isVDSOLoadModule(lm_nm)) {        
-        string path = "";
-        if (!prof.directorySet().empty()) {
-          // Assume we can find all version of the vdso in all measurement directories.          
-          // Can different runs encounter different vdso?
-          // If they can, it means a particular version of vdso may only
-          // show up in some of the measurement directories (not all).
-          // Then we cannot just choose the first measurement directory.                    
-          path = *(prof.directorySet().begin());
-        }        
-        path += "/" + lm_nm;
-        lm->open(path.c_str());
-      } else {
-        lm->open(lm_nm.c_str());
-      }
+      lm->open(lm_nm.c_str());
       lm->read(prof.directorySet(), BinUtil::LM::ReadFlg_Proc);
-
       if (vmaVec == NULL) {
 	DIAG_WMsgIf(printProgress, "Unable to compute struct simple for " << lm_nm);
       }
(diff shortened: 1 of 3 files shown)


mxz297:
@blue42u Can you give it a run on your machine to ensure that it works fine? It works fine on my end, but want to make sure.

blue42u:
Works like a charm on my end too.

I find it mildly amusing that all you did was use the same function the load modules use.

mxz297:
@mxz297 Right. It is the same normalization mechanism. Previously, I only applied it to Files. This time I apply it to also load modules and procedures.

mxz297:
Pushed a commit that add the virtual address and load module for a unknown procedure when reading from a structure file.

blue42u:
Got a run of Deepwave with this last patch, [database](https://rice.box.com/s/elbxdmy9m2wvi78imavhc9yc2nxajqol), [measurements](https://rice.box.com/s/ugwziimczi95o1bv03u2bs3xtki0qh5i) and [structs](https://rice.box.com/s/v2ltsz4wko3axrl45t5250o30vwejmlc). It looks like this:
image
Two issues (hopefully with easy fixes) come to mind:
 1. `<partial call paths>` shouldn't have the extra tags.
 2. When the load module is unknown we also shouldn't have the extra tags (for debugging partials we can use the logs). Maybe use `<unknown load module>` as the function name to distinguish it.

mxz297:
@blue42u Can you try this patch:

```
diff --git a/src/lib/analysis/Util.cpp b/src/lib/analysis/Util.cpp
index ea6154197..447804289 100644
--- a/src/lib/analysis/Util.cpp
+++ b/src/lib/analysis/Util.cpp
@@ -340,12 +340,16 @@ demandStructure(VMA vma, Prof::Struct::LM* lmStruct,
     else {
       char tmp_buf[TMP_BUFFER_LEN];
       string unknown_proc = (unknownProcNm) ? *unknownProcNm : string(UNKNOWN_PROC);
-
-      snprintf(tmp_buf, TMP_BUFFER_LEN, " 0x%lx", vma);
-      unknown_proc += tmp_buf;
-      if (lmStruct != NULL) {
-        snprintf(tmp_buf, TMP_BUFFER_LEN, " [%s]", FileUtil::basename(lmStruct->name().c_str()).c_str());
-        unknown_proc += tmp_buf;
+      if (unknown_proc == UNKNOWN_PROC) {
+        snprintf(tmp_buf, TMP_BUFFER_LEN, " 0x%lx", vma);
+        unknown_proc += tmp_buf;        
+        if (lmStruct != NULL) {
+          std::string base = FileUtil::basename(lmStruct->name().c_str());
+          if (base != UNKNOWN_LOAD_MODULE) {
+            snprintf(tmp_buf, TMP_BUFFER_LEN, " [%s]", base.c_str());
+            unknown_proc += tmp_buf;
+          }
+        }
       }
       Struct::File * fileStruct = Struct::File::demand(lmStruct, UNKNOWN_FILE);
       Struct::Proc * procStruct = Struct::Proc::demand(fileStruct, unknown_proc);
```

blue42u:
That works, but it leaves the `0x` tag on the `<unknown procedure>`'s.
image

I'm not certain if they should be kept or not, my preference is not (to a dev, virtual addresses after the program is dead is just spam). Any opinion @jmellorcrummey?

mxz297:
@blue42u  Oh. John specifically wants these virtual addresses and load modules for stripped binaries. I think "unknown load module" is not really useful. 

@jmellorcrummey So maybe we should not add virtual address and load module when the load module is unknown?

blue42u:
The stripped ones look fine, it makes sense to keep the offset for those:
image
When its unknown hpcrun will output the virtual address, which doesn't help after the program is dead (except maybe for unwind debugging).

When you have a thousand of these, like with Deepwave, it gets a little hard to find anything. But that's my opinion on the matter.

mxz297:
@blue42u I just have a chat with John. He believes that "unknown load module" means completely wrong unwind, and we should get rid of them completely. He thinks we should figure out the root cause rather than worry about the format of trash data. I think you can create a new issue for this, or maybe this is part of the unwind issue you are dealing with.

blue42u:
I'd love to get rid of them completely. We already have an issue for one (#213), I'll make a checklist issue over the weekend with any others I can shake out.

I think we have to worry about the format of trash data given that perfect data is nonviable, but this one's not a blocker for the release. I'll argue it further next week after things have settled down.

PR #216, opened 2020-02-18
Suppress samples during dlopen and dlclose

mxz297:
Use a thread local counter to suppress samples during dlopen and dlclose

### DIFF
--- a/src/tool/hpcrun/sample-sources/itimer.c
+++ b/src/tool/hpcrun/sample-sources/itimer.c
@@ -210,6 +210,7 @@ static __thread bool wallclock_ok = false;
  * external thread-local variables
  *****************************************************************************/
 extern __thread bool hpcrun_thread_suppress_sample;
+extern __thread int hpcrun_thread_dl_operation;
 
 // ****************************************************************************
 // * public helper function
@@ -690,7 +691,7 @@ itimer_signal_handler(int sig, siginfo_t* siginfo, void* context)
   sample_source_t *self = &_itimer_obj;
 
   // if sampling is suppressed for this thread, restart timer, & exit
-  if (hpcrun_thread_suppress_sample || sample_filters_apply()) {
+  if (hpcrun_thread_dl_operation > 0 || hpcrun_thread_suppress_sample || sample_filters_apply()) {
     TMSG(ITIMER_HANDLER, "thread sampling suppressed");
     hpcrun_restart_timer(self, 1);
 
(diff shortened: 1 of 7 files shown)


mxz297:
@jmellorcrummey @martyitz @mwkrentel  Using an integer counter can guard against recursive dlopen comparing to using a boolean. I think this represents a better version to try with SPARC.

Jokeren:
@mxz297 , the index values for `dlopen` and `dlclose` can be retrieved here.

https://github.com/HPCToolkit/hpctoolkit/blob/master/src/tool/hpcrun/hpcrun_flag_stacks.c#L7

Jokeren:
There was one shortcoming of our original design. We suppose an application will not call `dlopen` recursively more than 32 times.

src/tool/hpcrun/sample-sources/itimer.c
```diff
@@ -690,7 +691,7 @@ itimer_signal_handler(int sig, siginfo_t* siginfo, void* context)
   sample_source_t *self = &_itimer_obj;
 
   // if sampling is suppressed for this thread, restart timer, & exit
-  if (hpcrun_thread_suppress_sample || sample_filters_apply()) {
+  if (hpcrun_thread_dl_operation > 0 || hpcrun_thread_suppress_sample || sample_filters_apply()) {
     TMSG(ITIMER_HANDLER, "thread sampling suppressed");
```
jmellorcrummey: replace the first two conditions with a call to hpcrun_suppress_sample()

mxz297:
@jmellorcrummey Glad that it is sufficient to make progress on sparc. My goal was to quickly implement your idea to test it. I will refactor the code accordingly tomorrow.

mxz297:
@jmellorcrummey Pushed a commit for refactoring.

mwkrentel:
Actually, there already is a thread-specific variable to track inside
dlopen/dlclose.  This is defined in `thread_data.h` and used in
`hpcrun_dlfns.c`.

```
// True if this thread is inside dlopen or dlclose.  A synchronous
// override that is called from dlopen (eg, malloc) must skip this
// sample or else deadlock on the dlopen lock.
bool inside_dlfcn;
```

This is used to protect the synchronous override of malloc() for
memleak, but not for dl_iterate_phdr() from libunwind.  (It predates
libunwind.)

This goes back to 2011, so I don't know if it handles the recursive
dlopen case correctly.

I can probably make either of them break by going nuclear with
longjmp().  (Whenever I see nested increment/decrement, I first think
of what happens if you long jump from somwhere inside.)  But I don't
know a reasonable way of handling longjmp from inside the application.

Anyway, I'm basically Ok with this patch.

PR #220, opened 2020-02-22
1. Add RELA relocation for .debug_line section; 2. Do not relocate fa…

Jokeren:
@jmellorcrummey , @mxz297 

Two problems are fixed in this commit.

1. Previously, our code intended to handle both cubins and fatbins. However, it appears that we handled cubins incorrectly. We first use `elfFile->open` to relocate the specified cubin, then in `findCubins` we get another cubin and relocate it again. 
  You may think we just relocate the cubin twice, but somehow, mysteriously, we relocate a cubin four times and cause a segment fault in the last time. I was not able to figure out why this happened. The bottom line is that we can stub out the code for fatbin as we can only handle separated cubins dumped by hpcrun.

https://github.com/HPCToolkit/hpctoolkit/blob/53ed9dd383dcd5dc908b033e9b717eda58fcc8b0/src/lib/binutils/InputFile.cpp#L165

https://github.com/HPCToolkit/hpctoolkit/blob/53ed9dd383dcd5dc908b033e9b717eda58fcc8b0/src/lib/binutils/Fatbin.cpp#L192

https://github.com/HPCToolkit/hpctoolkit/blob/53ed9dd383dcd5dc908b033e9b717eda58fcc8b0/src/lib/binutils/Fatbin.cpp#L153

2. Optimized cubins usually only contain a `.debug_line` section with `REL` relocation info. However, we have got a cubin---not sure how it was compiled, that has a `.debug_line` section with `RELA` relocation. Therefore, I added another path to deal with it.

### DIFF
--- a/src/lib/binutils/RelocateCubin.cpp
+++ b/src/lib/binutils/RelocateCubin.cpp
@@ -247,6 +247,26 @@ applyLineMapRelocations
 }
 
 
+static void
+applyDebugInfoRelocations
+(
+ Elf_SymbolVector *symbol_values,
+ char *debug_info,
+ int n_relocations,
+ Elf_Data *relocations_data
+)
+{
+  //----------------------------------------------------------------
+  // apply each debug info relocation
+  //----------------------------------------------------------------
+  for (int i = 0; i < n_relocations; i++) {
+    GElf_Rela rela_v;
+    GElf_Rela *rela = gelf_getrela(relocations_data, i, &rela_v);
+    applyRELArelocation(debug_info, symbol_values, rela);
+  }
+}
+
+
 // if the cubin contains a line map section and a matching line map relocations
 // section, apply the relocations to the line map
 static void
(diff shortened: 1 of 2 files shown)


src/lib/binutils/RelocateCubin.cpp
```diff
@@ -247,6 +247,26 @@ applyLineMapRelocations
 }
 
 
+static void
+applyDebugInfoRelocations
```
jmellorcrummey: Since we are no longer using these just for DebugInfo, the name should change to 

applyRELArelocations

src/lib/binutils/RelocateCubin.cpp
```diff
@@ -319,8 +340,13 @@ relocateLineMap
 #endif
               if (n_relocations > 0) {
                 Elf_Data *relocations_data = elf_getdata(scn, NULL);
-                applyLineMapRelocations(symbol_values, line_map,
-                  n_relocations, relocations_data);
+                if (shdr.sh_type == SHT_RELA) {
+                  applyDebugInfoRelocations(symbol_values, line_map,
+                    n_relocations, relocations_data);
+                } else {
+                  applyLineMapRelocations(symbol_values, line_map,
```
jmellorcrummey: the name of applyLineMapRelocations should also be changed to applyRELrelocations

Jokeren:
@jmellorcrummey , function names are changed.

PR #230, opened 2020-02-25
Ppc64 unwinder tail call

jmellorcrummey:
Add support for recognizing interior tail calls when creating PPC64 unwind recipes.

@mwkrentel @mxz297 please review the code and test further. the memstress test now has no errant unwinds. 

quicksilver works too, though I now notice 2 unknown procedures. these may have been a consequence of keren's recent change (but not this commit).


mxz297:
@jmellorcrummey I ran this PR with my GPU example with a real timer. All samples are fine.

mwkrentel:
Yes, this also works for me.

PR #233, opened 2020-02-27
Connect dangling blocks only when instruction offsets are not duplicate

Jokeren:


mxz297:
@jmellorcrummey @Jokeren 

I pulled in this PR locally and it looks to me that the inconsistent CUBIN CFG issue is resolved. However, I got the following figure:

image

You can see we don't have unknown procedures on the GPU side but we got unknown procedures on the CPU side. The annotated virtual addresses for the unknown procedures are not 4-byte aligned. I vaguely remember that Lai's trampoline code will subtract the return address by 1, but the trampoline code should be disabled by default. 

Is there any other situation where we subtract the return address by 1?

PR #241, opened 2020-03-12
Force libunwind to not call dl_iterate_phdr during unwind

mxz297:
The first goal of this PR is to prevent the deadlock caused by the lock used by `dlopen`, `dlsym`, `dlclose`, and `dl_iterate_phdr`. The overall approach is to maintain our own load map and feed it to libunwind, so that libunwind does not call `dl_iterate_phdr`. The approach is to intercept calls to `dl_iterate_phdr` in libunwind with our own version of `dl_iterate_phdr` using GOTCHA. 

The second goal is to refactor the use of dlopen/dlsym to identify modules to ignore. I use libelf to read the .dynsym sections to find function names. 

The third goal is to clean up redundant code that deals with library loading operations.

@jmellorcrummey @Jokeren @mwkrentel This PR is locally tested myself with Mark's dlstress unit test and some small nvidia example. It still needs several improvement before merging, but I would like to start to get comments. A few items for comments:

1. Currently, you need to use my fork of GOTCHA (https://github.com/mxz297/GOTCHA/tree/library_filter). Note that currently there is no public interface in GOTCHA to only rewrite .got for a particular module. What kind of interface do we want?  Do we want to only rewrite .got for a certain module? Do we want to extend this a little bit more so that we can control which .got entry in which module to overwrite?

2. GOTCHA is currently not a dependency of HPCToolkit. Do we want to use this as an opportunity to officially add GOTCHA as a dependency? And how should we do this?

3. @Jokeren What's the purpose of the flag stack (https://github.com/HPCToolkit/hpctoolkit/blob/master/src/tool/hpcrun/hpcrun_flag_stacks.c)? Is this stack needed for using dlopen/dlsym to check whether a module contains CUDA functions? If I have replaced that with using libelf, is it safe to remove this code?

4. Do we still need the reader/writer lock used during stack unwinding? The writer lock was used to protect modifying our load map. The reader lock was used to protect taking samples (https://github.com/HPCToolkit/hpctoolkit/blob/master/src/tool/hpcrun/sample_event.c#L199). In our code, we say the reader locker will block when taking a synchronous sample, but will give up when taking an asynchronous sample. Do we still need to distinguish sync and async samples?

5. What's the factors to consider for static binaries? Can statically linked executable perform dynamic loading?


Jokeren:
The flag stack is used to prevent hpcrun from crashing in recursive dlopen cases. There are two points here. 
1. The use of flag. If for some reasons, hpcrun is not initialized or not able to enter a safe region, we do not call `monitor_pre_dlopen`. And because `hpcrun_pre_dlopen` is not called, neither do we call `monitor_dlopen` . 
2. The use of stack. When a dlopen is used inside a dlopen, a single flag is not enough to indicate whether a `monitor_dlopen` call's correponding `monitor_pre_dlopen` succeeded or not.

mxz297:
@jmellorcrummey @mwkrentel Since my change to gotcha has been merged and there is already a new tag that includes my change (https://github.com/LLNL/GOTCHA/tree/v1.0.3), I updated my code accordingly. I have also done some code cleaning to get rid of the reader/writer lock and the dlopen flag stack. 

I have tested the new code with the dl stress tests without seeing any regression.

mwkrentel:
@mxz297 Does the new version plus new gotcha work with the dlstress
test with dlsym()?  I'll give it a go and see if I can find even harder cases.

I'll also update the spack recipe for gotcha, unless Matt beats me to it.

mxz297:
> @mxz297 Does the new version plus new gotcha work with the dlstress
> test with dlsym()? I'll give it a go and see if I can find even harder cases.

@mwkrentel Yes, working with dlsym test. Will be good to find even harder cases.
> 
> I'll also update the spack recipe for gotcha, unless Matt beats me to it.

mwkrentel:
I've started looking at this.

1. I'll want to update the configure test to make it more robust and
add rpath.

2. The static case is broken because you can't link gotcha into
libhpcrun.o.  But that's easy to fix.

3. I've updated the spack recipe for 1.0.3, but maybe you're going to
need 1.0.4.

4. Your branch works with my dlopen() + dlsym() test, which is good.
(Previously we had deadlock with dlsym.)  I'll see if I can make some
harder but legitimate tests.

mxz297:
> 3. I've updated the spack recipe for 1.0.3, but maybe you're going to
>    need 1.0.4.

GOTCHA 1.0.4 will be needed when we start to work with Intel OpenCL, which means after merging Aaron's code. John and I will review Aaron's code on Thursday. We will then know whether OpenCL support will be good to merge.

mxz297:
Closed via #276

PR #244, opened 2020-03-16
Fnbounds

martyitz:
Finish eh_frame support


jmellorcrummey:
lightly tested prior to rebasing and found it to work fine.

PR #246, opened 2020-03-21
Cupti buffer fix

Jokeren:
Use `finalize_event_list` to register Cupti activities without knowing CUDA context. 

Only PC sampling is registered when a context is created because pc sampling configuration needs to know which context the kernel belongs to. 

Tests for Laghos, Quicksolver, and simpleCooperativeGroups have been done. 

@mxz297 , can you double check this commit?

### DIFF
--- a/src/tool/hpcrun/sample-sources/nvidia.c
+++ b/src/tool/hpcrun/sample-sources/nvidia.c
@@ -122,6 +122,7 @@ static device_finalizer_fn_entry_t device_trace_finalizer_shutdown;
 // default trace all the activities
 // -1: disabled, >0: x ms per activity
 // static long trace_frequency = -1;
+static long trace_frequency = -1;
 static long trace_frequency_default = -1;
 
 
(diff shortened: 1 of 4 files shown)


mxz297:
Looking good to me.

Jokeren:
@jmellorcrummey , can we merge this branch?

src/tool/hpcrun/sample-sources/nvidia.c
```diff
@@ -412,8 +404,8 @@ METHOD_FN(process_event_list, int lush_metrics)
 
   // Register cupti callbacks
   cupti_init();
-  cupti_callbacks_subscribe();
   cupti_start();
```
jmellorcrummey: Why do we subscribe after starting cupti? Is this an equivalent choice? If so, I prefer subscribe before start. If not equivalent, it needs a comment.

jmellorcrummey: Other than my comment, this looks good. One other question: is the prior race eliminated or just less likely?

Jokeren: Oh, it does not matter where to put cupti_callbacks_subscribe and cupti_start. They are equivalent.

Jokeren: The prior race is eliminated completely as far as I can tell.

Issue #248, opened 2020-03-23
hpcprof-mpi -M stats fails for GPU events


laksono:
Platform: gpu.cs
Branch: master

Profiling a CUDA application with `-e gpu=nvidia,pc` and then try to generate the database using `-M stats` causes an exception. This works on other events except GPU events. This works fine if we generate the database **without**  `-M stats`.

```
$ mpirun -np 2 hpcprof-mpi -M stats -S simpleMultiGPU.hpcstruct  hpctoolkit-simpleMultiGPU-measurements/
msg: STRUCTURE: /home/la5/pkgs-src/cuda-samples/0_Simple/simpleMultiGPU/simpleMultiGPU
msg: Line map : /tmp/pkgs/hpctoolkit-cleanmaster/lib/hpctoolkit/libhpcrun.so.0.0.0
msg: Line map : /tmp/pkgs/hpctoolkit-cleanmaster/lib/hpctoolkit/ext-libs/libmonitor.so.0.0.0
msg: Line map : /usr/lib64/libcuda.so.418.40.04
msg: Line map : /usr/local/cuda-10.0/extras/CUPTI/lib64/libcupti.so.10.0.130
msg: Line map : /home/la5/pkgs-src/cuda-samples/0_Simple/simpleMultiGPU/hpctoolkit-simpleMultiGPU-measurements/cubins/9945797cf35e8954f93b4003a65abc18.cubin
msg: Populating Experiment database: /home/la5/pkgs-src/cuda-samples/0_Simple/simpleMultiGPU/hpctoolkit-simpleMultiGPU-database
ERROR: [Diagnostics::FatalException] Unexpected input:  [../../../../src/lib/prof/CallPath-Profile.cpp:832]
```
The error is triggered by an exception at line 832:
```
 816       // 0. retrieve combine formula (each DerivedIncrDesc corresponds
 817       // to an 'accumulator')
 818       string combineFrm;
 819       if (mDrvdExpr) {
 820         combineFrm = mDrvdExpr->combineString1();
 821
 822         for (uint k = 1; k < mDrvdExpr->numAccum(); ++k) {
 823           uint mId = mDrvdExpr->accumId(k);
 824           string frm = mDrvdExpr->combineString2();
 825           metricIdToFormula.insert(std::make_pair(mId, frm));
 826         }
 827       }
 828       else {
 829         // must represent accumulator 2
 830         uint mId = m->id();
[...]

Issue #66, opened 2018-03-23
bad configure test for perf events on blue gene


mwkrentel:
There's a problem with configure and perf events in mult-kernel on
blue gene.  Blue gene doesn't support perf events (and even crashes)
but configure runs front-end tests and concludes that it does.

This is a hard problem because you really need to run a test on the
compute node.  Or at least treat this as a cross compile and use a
back-end compiler, which we don't do.

There is a similar problem with REALTIME where the front end compiler
passes all the tests, but the code doesn't run on the back end.

Anyway, we already run similar tests in externals for a thing in
libmonitor and in toolkit for turning off REALTIME.  So, I can just
adapt those tests to turn off perf events for blue gene.

I just want a little time to think about the cleanest way to set the
configure options so that it's still possible to build for the front
end, if desired.

I'll set this priority one because the default directions that work
for master will crash at runtime on blue gene.

mwkrentel:
Now fixed in mult-kernel in commit f0cb7ca828 on 2018-04-02.

I added the option --enable-bgq and auto-detect blue gene and build
for the back end by default.  This turns off the POSIX timers and Perf
Events (which were generating an illegal instruction).

This can now be closed because it was only broken in mult-kernel.
It was never broken in master (doesn't have perf events yet).

Issue #16, opened 2017-09-25
<program root> and <thread root> should not clickable in hpcviewer


laksono:
In hpcviewer, we need to avoid users to click <program root> and <thread root> nodes and showing the source code of libmonitor. These special nodes should be unclickable. Either hpcviewer needs to "filter" the nodes or hpcrun or hpcprof has to be smart  enough to avoid linking the special nodes to the source files:

    <File i="35086" n="./src/home/la5/git/hpctoolkit-externals/libmonitor/libmonitor/src/pthread.c"/>
    <File i="35089" n="./src/home/la5/git/hpctoolkit-externals/libmonitor/libmonitor/src/main.c"/>
    <File i="35092" n="./src/home/la5/git/hpctoolkit-externals/libmonitor/libmonitor/src/mpi_init_c.c"/>

jmellorcrummey:
I believe that these nodes only have hyperlinks when libmonitor.so is not stripped. It is normally stripped (for this exact reason) unless you compile with --enable-develop.

Issue #15, opened 2017-09-11
hpcrun needs to record meta-data in hpcrun files


jmellorcrummey:
hpcrun needs to record meta-data about the execution including the node id and thread id (or thread ids) it ran on.

laksono:
I recommend to record also the command line. Sometimes we forget the command line to produce the measurements.

blue42u:
[This repository has moved to GitLab, please continue this issue there.](https://gitlab.com/hpctoolkit/hpctoolkit/-/issues/15)

PR #249, opened 2020-03-24
Fix metric stats

laksono:
Fixing issue #248 (-M stats fails for GPU events)

**Symptom**: 
  hpcprof and hpcprof-mpi throw an exception when generating a database for GPU-based events

**Cause**: 
  Conflict of using formula in hpcprof.   Inside hpcprof, we always copy metric formula from the original metric to the derived metrics. This includes statistic derived metrics like Mean, StdDev, Max, Min, Cov, etc. These metrics have their own expression and formula, and shouldn't be mixed with metrics with hpcrun's formula.

**Solution**:
  Copy the formula iff the derived metric is from Sum-type operator.


laksono:
Just to clarify, at the moment hpcviewer cannot handle if a metric is both hpcrun-derived metric AND statistical metric (Mean, Max, Min, ...). 

This database can be handled by hpcviewer (but the output of metric Mean will be empty):
```
    <Metric i="554" n="GSAMP:UTIL:Sum (I)" o="18" v="derived-incr" ...>
      <MetricFormula t="view" frm="min(100, max(0, 100*#16/#15))"/>
      <Info><NV n="units" v="events"/></Info>
    </Metric>
    <Metric i="555" n="GSAMP:UTIL:Mean (I)" o="-1" v="derived-incr" ...>
      <MetricFormula t="combine" frm="sum($555, $555)"/>
      <MetricFormula t="finalize" frm="$555 / $556"/>
      <Info><NV n="units" v="events"/></Info>
    </Metric>
```

But if hpcprof generates both hpcrun-derived formula (see the attribute "view") and derived incremental formula (attributes "combine" and "finalize"), this confuses hpcviewer when parsing "`GSAMP:UTIL:Sum (I)`" metric:

```
    <Metric i="554" n="GSAMP:UTIL:Sum (I)" o="18" v="derived-incr" ...>
      <MetricFormula t="view" frm="min(100, max(0, 100*#16/#15))"/>
      <MetricFormula t="combine" frm="sum($554, $554)"/>
      <MetricFormula t="finalize" frm="$554"/>
      <Info><NV n="units" v="events"/></Info>
    </Metric>
    <Metric i="555" n="GSAMP:UTIL:Mean (I)" o="-1" v="derived-incr" ...>
      <MetricFormula t="combine" frm="sum($555, $555)"/>
      <MetricFormula t="finalize" frm="$555 / $556"/>
      <Info><NV n="units" v="events"/></Info>
    </Metric>
```
I am working on hpcprof (to allow generating both formula) and hpcviewer to be able to handle this.

laksono:
To check this branch:
- Profile a GPU application with GPU events like: `-e gpu=nvidia,pc`
- Run hpcprof or hpcprof-mpi using `-M stats` flag
- If it doesn't crash, there it's good.
  - The statistic metric values for hpcrun derived metrics (like GSAMP:UTIL) are empty since hpcviewer doesn't know how to deal with the mix of derived metrics and statistic metrics.

jmellorcrummey:
rather than having empty stats metrics for derived metrics, it would be better to not define them and to just say you don’t get them for derived metrics. or, if stats are requested, the derived metrics defined by formulae could be computed by prof before stats are computed.

laksono:
> rather than having empty stats metrics for derived metrics, it would be better to not define them and to just say you don’t get them for derived metrics. or, if stats are requested, the derived metrics defined by formulae could be computed by prof before stats are computed.

I meant the values for `Sum` metric are correct. However, for `Mean`, `Max`, `Min`, etc the values are wrong since hpcprof is unable to generate combine and finalize formula that suit hpcviewer.

For example, in this experiment.xml below, hpcviewer knows how to compute `Sum` metric since metrics `#16` and `#15` have raw values generated by hpcprof.

However, hpcviewer doesn't know how to compute metric `Mean` because hpcprof cannot generate its raw value (hpcprof always generates raw values for stats metrics). Since there are no raw values, hpcviewer cannot compute the formula `sum($555, $555)`. That's why for these metrics, the columns are empty.
``` 
   <Metric i="554" n="GSAMP:UTIL:Sum (I)" o="18" v="derived-incr" ...>
      <MetricFormula t="view" frm="min(100, max(0, 100*#16/#15))"/>
      <Info><NV n="units" v="events"/></Info>
    </Metric>
    <Metric i="555" n="GSAMP:UTIL:Mean (I)" o="-1" v="derived-incr" ...>
      <MetricFormula t="combine" frm="sum($555, $555)"/>
      <MetricFormula t="finalize" frm="$555 / $556"/>
      <Info><NV n="units" v="events"/></Info>
    </Metric>
```
Btw, we agreed that hpcrun's derived metrics will be computed by hpcviewer, not hpcprof. The reason is that if they are computed by hpcprof, it will show incorrect values in bottom-up and flat trees.
As I proposed in hpcviewer issue `#66` (https://github.com/HPCToolkit/hpcviewer/issues/66), it's much easier if we don't generate stats metrics for derived metrics.

jmellorcrummey:
I agree that we shouldn't generate stats for derived metrics for now.

jmellorcrummey:
Since nothing in the pull request seems to change what is computed, I assume that this branch just causes empty columns for stats to be inserted. Rather than do that, it would be better to just eliminate the empty columns of uncomputed stats.

laksono:
This patch avoids propagating hpcrun derived metrics to statistic metrics other than Sum.
Hence, we just compute the derived metrics only from Sum, while others are empty. This patch is to avoid a crash while running hpcprof.

In other branch (fix-metric_name), by default hpcrun removes statistic metrics in derived metrics. 
My preference is to remove them, but you argued we may need to keep them (see the discussion in hpcviewer issue #66)

PR #252, opened 2020-04-02
Remove --metric-db from hpcprof

laksono:
hpcprof at the moment doesn't support --metric-db, we should remove this option from the help (-h) and the man page.


jmellorcrummey:
Looks fine to me.

PR #254, opened 2020-04-06
CUPTI Debug Logging

blue42u:
Changes the CUDA logging from using compile-time `printfs` to using Run's debug logging system. Because I got tired of recompiling every time.

Probably the tags are wrong, I'll take changes if anyone (esp. @Jokeren) has a better idea.


Jokeren:
The changes are ok.

PR #265, opened 2020-04-25
Noread exec

jmellorcrummey:


jmellorcrummey:
when an executable is unreadable, hpcfnbounds can't provide function bounds for functions in the executable. in fact, hpctoolkit failed to even identify the address range associated with the executable and attributed samples to [unknown load module]. 

This change uses dl_iterate_phdr and treats the load module with the name "" as the executable. No documentation says this is right, but the bounds associated with load module "" are the bounds for the executable. 

Add support to dylib to find bounds for a load module by name. find bounds for an executable by asking for bounds for the load module named "". ask dylib to compute bounds for an executable if hpcfnbounds doesn't yield any useful information.

The test case for this is hpctoolkit-tests/applications/fib-noread.

mwkrentel:
I must be missing something.  I tried your fib-noread program with
master and the noread-exec branch, both with fnbounds2.  (You missed
the switch by 1 commit, but I can set it at the command line.)

In both cases, I get a useless profile with 100% partial unwinds, all
'unknown procedure', and all errors in the log file.

```
SAMPLE ANOMALIES: blocks: 0 (async: 0, dlopen: 0), errors: 3257 (segv: 0, soft: 3257)
SUMMARY: samples: 3257 (recorded: 3257, blocked: 0, errant: 3257, trolled: 0, yielded: 0),
```

The only difference I see is the load module says 'fib' in the new
profile and 'unknown load module' in the old.  Is that the whole
point?

Also, hpcprof in the noread branch generates this error, I assume from
struct simple.

```
ERROR: unable to open input file: /home/krentel/noread/fib
WARNING: Cannot fully process samples for load module /home/krentel/noread/fib:
'/home/krentel/noread/fib': Permission denied
msg: Line map : /home/krentel/noread/libfib.so
msg: Populating Experiment database: /home/krentel/noread/database-noread
```

So, a similar adjustment needs to be made in hpcprof.

----------

The fib/libfib build seems a bit convoluted.  fib calls libfib and
passes a function argument, so the two call each other back and forth.
Also, fib is set unreadable, but not libfib.so.

Is this necessary to replicate the conditions at LLNL?

Would it suffice to have fib call libfib, stay in libfib and set
libfib unreadable?  Or even, a single fib binary that is executable
and not readable?

----------

This patch, when ready is a good candidate to rebase instead of
merging.  I'm getting tired of this double helix in master.  Makes
things hard to track down.

I can show you what to do and why that's adventageous.  But don't merge
master into noread-exec, that will just complicate things.

mwkrentel:
Uh, I would hold off on merging until at least the hpcprof issue is fixed.

mxz297:
> Uh, I would hold off on merging until at least the hpcprof issue is fixed.

@mwkrentel Ah. I probably misused this review feature. The current code changes look good to me, but obviously as you reported, more changes are needed.

jmellorcrummey:
> I must be missing something. I tried your fib-noread program with master and the noread-exec branch, both with fnbounds2. (You missed the switch by 1 commit, but I can set it at the command line.)

> In both cases, I get a useless profile with 100% partial unwinds, all 'unknown procedure', and all errors in the log file.

> The only difference I see is the load module says 'fib' in the new profile and 'unknown load module' in the old. Is that the whole point?

Yes. When running code at the NNSA labs, occasionally one encounters unreadable binaries. Rather than saying 'it is unknown where these samples came from' we should say "the samples came from 'name-of-unreadable-binary' ". Oh, by the way, we can't unwind through the unreadable binary because we don't have eh_frame info for libunwind and we can't do on-the-fly binary analysis because the binary is unreadable.

> The fib/libfib build seems a bit convoluted. fib calls libfib and passes a function argument, so the two call each other back and forth. Also, fib is set unreadable, but not libfib.so.

I wanted to make sure that I had all of the support that I needed. I first handled the unreadable executable. Then, I tried creating libfib.so as an unreadable binary. It turns out that one can't use an unreadable shared library. Apparently, the loader demands to read some of its information.

> Is this necessary to replicate the conditions at LLNL?

Yes.

> Would it suffice to have fib call libfib, stay in libfib and set libfib unreadable? Or even, a single fib binary that is executable and not readable?

The case we ran into at LLNL was an unreadable binary that called readable shared libraries. A user wanted to profile unreadable ALE3D that called his readable library. This test replicates the essential details.

> hpcprof on the fib-noread example throws the following error.

```
ERROR: unable to open input file: /home/krentel/noread/fib
WARNING: Cannot fully process samples for load module /home/krentel/noread/fib:
'/home/krentel/noread/fib': Permission denied
```

The improved error/warning message is:

```
WARNING: Cannot fully process samples for load module /home/johnmc/release-planning/hpctoolkit-tests/applications/fib-noread/fib: unable to open input file /home/johnmc/release-planning/hpctoolkit-tests/applications/fib-noread/fib (Permission denied)
```

mwkrentel:
Ok.  Is there any reason to keep the exec_wait variable?
It generates a warning for unused variable.

```
int 
dylib_find_executable_bounds(void** start, void** end)
{
  static volatile int exec_wait = 1;

  // executable name in map is empty string; don't know why
  return dylib_find_module_bounds_by_name("", start, end);
}
```

mwkrentel:
Grrr.  You continued to do the merge double helix.  :-(
This makes it harder to track down where problems are introduced.

I can show you how to avoid this with rebase.

PR #267, opened 2020-05-06
Add GPU and OMP place folders as "fake" procedures

laksono:
This is to help hpcviewer to distinguish between place folders and normal procedures. 
hpcviewer will add the name of the load module if a node doesn't have the source code. However,
if a node is a place holder, hpcviewer should not add the load module name (since the source code is not available).


jmellorcrummey:
Perfect. Thanks for making this adjustment.

PR #270, opened 2020-05-15
Disable cubin relocation in hpcprof

Jokeren:


src/lib/binutils/LM.cpp
```diff
@@ -411,6 +411,10 @@ BinUtil::LM::open(const char* filenm)
   }
 
   // Write relocated cubins and reopen them
+  // TODO(Keren): fix the following problems before dumping relocate cubins
```
jmellorcrummey, line 414: I think that the correct fix is more subtle than this. This code reads a cubin, writes a relocated cubin, and then changes filenm to point to the relocated file. Without this code, the line map code will try to open the original cubin and read the line map. this will cause chaos because all of the functions are at address 0 and their line maps overlap.

I think the right fix is to try opening the file and if it is a cubin return with an error rather than letting the code fall through and try to open the original cubin file.

Jokeren: But fix this does not solve all the problems. We also have to avoid races in the filesystem.

jmellorcrummey: The race in the filesystem was one thread trying to write a .relocate while another thread is trying to read a .relocate file. If we take out the write of .relocate, there is no race in the filesystem. 

As I mentioned above, falling through and trying to read the cubin linemap will cause us other problems that weren't visible because of the filesystem race.

Jokeren: I'll need to think about it. I think what you said should be one *process* trying to write a relocate file while when another *process* is reading the relocate file, right?

jmellorcrummey: Right. I think that was a problem with hcpprof-mpi. I don't think we saw it before with small cubins. At Sandia, they have an immense cubin for adagio, which increases the likelihood of a race causing a problem.

Jokeren:
@jmellorcrummey I let hpcprof print a warning message if a cubin is opened without a corresponding structure file.

jmellorcrummey:
this has been superseded by changes to master.

PR #282, opened 2020-06-08
Master gpu trace

dejangrubisic:
Created


src/tool/hpcrun/control-knob.c
```diff
@@ -30,20 +34,23 @@ void
...
 #define INIT_KNOBS(knob_name) \
   control_knobs[knob_name].name = control_knob_names[knob_name]; \
-  control_knobs[knob_name].value = NULL;
+  control_knobs[knob_name].value = default_knob[knob_name];
 
   FORALL_KNOBS(INIT_KNOBS) 
 
 #undef INIT_KNOBS
 
+	if (s) {
```
jmellorcrummey, line 46: A separate issue: at least one of these control knobs (related to semaphores)  is nvidia specific and shouldn't be available if we are not using an NVIDIA GPU.

src/tool/hpcrun/main.c
```diff
@@ -685,10 +686,11 @@ hpcrun_fini_internal()
       return;
     }
 
-    // Call all registered auxiliary functions before termination.
-    // This typically means flushing files that were not done by their creators.
-    device_finalizer_apply(device_finalizer_type_flush);
-    device_finalizer_apply(device_finalizer_type_shutdown);
+    //comment this dejan
```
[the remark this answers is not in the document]
Jokeren: These are commented probably when I was debugging the data race bug. Should be denoted as `uncomment this dejan`.

src/tool/hpcrun/main.c
```diff
@@ -1159,13 +1186,13 @@ monitor_init_thread(int tid, void* data)
   void *thread_begin_address = monitor_get_addr_thread_start();
 
   if (module_ignore_map_inrange_lookup(thread_begin_address)) {
-    hpcrun_thread_suppress_sample = true;
+		hpcrun_thread_suppress_sample = true;
```
Jokeren, line 1189: Indentation

src/tool/hpcrun/sample-sources/nvidia.c
```diff
@@ -397,14 +392,7 @@ METHOD_FN(process_event_list, int lush_metrics)
...
     control_knob_value_get_int(HPCRUN_CUDA_DEVICE_SEMAPHORE_SIZE);
 
-  if (device_buffer_size == 0) {
-    device_buffer_size = DEFAULT_DEVICE_BUFFER_SIZE;
-  }
-
-  if (device_semaphore_size == 0) {
-    device_semaphore_size = DEFAULT_DEVICE_SEMAPHORE_SIZE;
-  }
-
```
Jokeren, line 407: These variables are nvidia specific. I would suggest to initialize the variables here.

src/tool/hpcrun/threadmgr.c
```diff
@@ -163,7 +163,9 @@ allocate_and_init_thread_data(int id, cct_ctxt_t* thr_ctxt)
   // ----------------------------------------
   // opening trace file
   // ----------------------------------------
-  hpcrun_trace_open(&(data->core_profile_trace_data));
+  if( has_trace ){
+		hpcrun_trace_open(&(data->core_profile_trace_data));
```
Jokeren, line 167: Indentation

src/tool/hpcrun/threadmgr.c
```diff
@@ -163,7 +163,9 @@ allocate_and_init_thread_data(int id, cct_ctxt_t* thr_ctxt)
   // ----------------------------------------
   // opening trace file
   // ----------------------------------------
-  hpcrun_trace_open(&(data->core_profile_trace_data));
+  if( has_trace ){
```
Jokeren, line 166: if (has_trace) {

src/tool/hpcrun/gpu/gpu-trace.c
```diff
@@ -395,16 +400,13 @@ gpu_trace_record
...
+	while (!atomic_load(&stop_trace_flag)) {
+    //getting data from a trace channel
+    gpu_trace_activities_process(thread_args->channel_id);
     gpu_trace_activities_await(thread_args);
   }
-
-  gpu_trace_activities_process(td, thread_args);
-
-  gpu_trace_stream_release(td);
+	gpu_trace_activities_process(thread_args->channel_id);
```
Jokeren, line 408: Indentation

src/tool/hpcrun/gpu/gpu-trace.c
```diff
@@ -250,7 +255,7 @@ gpu_trace_start_adjust
 }
 
 
-static void
+void
```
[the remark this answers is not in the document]
Jokeren: This function is invoked in `gpu-trace-channel.c`.

src/tool/hpcrun/control-knob.h
```diff
@@ -4,6 +4,7 @@
 #define FORALL_KNOBS(macro)  \
   macro(HPCRUN_CUDA_DEVICE_BUFFER_SIZE)     \
   macro(HPCRUN_CUDA_DEVICE_SEMAPHORE_SIZE)  \
+	macro(STREAMS_PER_THREAD)	\
```
Jokeren, line 7: Indentation

src/tool/hpcrun/gpu/gpu-trace-channel.h
```diff
@@ -53,14 +56,19 @@
...
 
-
-
 //******************************************************************************
 // type declarations
 //******************************************************************************
 
 typedef struct gpu_trace_channel_t gpu_trace_channel_t;
 
+typedef struct gpu_trace_channel_t {
```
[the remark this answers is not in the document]
dejangrubisic: the problem is that that I am using **gpu_trace_channel_t** data structure in :
gpu-trace.c
gpu-trace-channel-set.c

and if I don't expose the abstraction in header file, it will not be included in these two files and I am getting :
error: dereferencing pointer to incomplete type ‘gpu_trace_channel_t {aka struct gpu_trace_channel_t}’

Is there a better way for this?

src/tool/hpcrun/gpu/gpu-trace-channel.c
```diff
@@ -167,24 +159,26 @@ gpu_trace_channel_produce
  gpu_trace_item_t *ti
 )
 {
-  gpu_trace_item_t *cti = gpu_trace_item_alloc(channel);
+  gpu_trace_item_t * cti = gpu_trace_item_alloc(channel);
```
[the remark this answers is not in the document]
dejangrubisic: done

src/tool/hpcrun/gpu/gpu-trace-channel.c
```diff
@@ -47,7 +47,7 @@
 
 #include <string.h>
 #include <pthread.h>
-
+#include "gpu-print.h"
```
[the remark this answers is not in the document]
dejangrubisic: got it

src/tool/hpcrun/control-knob.c
```diff
@@ -30,20 +34,23 @@ void
 control_knob_init()
 {
   char *s = getenv("HPCRUN_CONTROL_KNOBS");
-  if (s) {
+
 #define INIT_KNOBS(knob_name) \
   control_knobs[knob_name].name = control_knob_names[knob_name]; \
-  control_knobs[knob_name].value = NULL;
+  control_knobs[knob_name].value = default_knob[knob_name];
```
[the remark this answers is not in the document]
dejangrubisic: Thanks for the advice. Done :)

src/tool/hpcrun/gpu/gpu-trace-channel-set.h
```diff
@@ -0,0 +1,87 @@
...
+// or otherwise) arising in any way out of the use of this software, even
+// if advised of the possibility of such damage.
+//
+// ******************************************************* EndRiceCopyright *
+
+#ifndef gpu_trace_channel_set_h
+#define gpu_trace_channel_set_h
+
+
+#define MAX_THREADS_CONSUMERS 256
```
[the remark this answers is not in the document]
dejangrubisic: Sure. But if that is in knobs, it becomes a dynamic array, so I had to add 

void **gpu_trace_channel_stack_alloc**(int size){
	gpu_trace_channel_stack = hpcrun_malloc_safe( size * sizeof(typed_stack_elem_ptr(gpu_trace_channel_ptr_t)));
}
, And to call it from **schedule_multi_threads()**
gpu_trace_channel_stack_alloc(	control_knob_value_get_int(MAX_THREADS_CONSUMERS) );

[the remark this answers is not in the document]
dejangrubisic: if I am using **hpcrun_malloc_safe**, how do I free that memory?

src/tool/hpcrun/gpu/gpu-trace-channel-set.h
```diff
@@ -0,0 +1,87 @@
...
+
+//******************************************************************************
+// interface operations
+//******************************************************************************
+
+void
+gpu_trace_channel_set_insert
+(
+ gpu_trace_channel_t *channel,
+ int thread_num
```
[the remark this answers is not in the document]
dejangrubisic: Done, now for indexing channel_sets we use set_index

src/tool/hpcrun/gpu/gpu-trace-channel-set.h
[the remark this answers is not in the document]
jmellorcrummey: I don't understand. There is a difference between a channel stack and a channel set. You only want to dynamically allocate a vector (once) that can hold a pointer to an array of pointers to MAX_THREADS_CONSUMERS channel stacks.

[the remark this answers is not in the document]
jmellorcrummey: If you use hpcrun_malloc_safe for this ONCE, you don't need to free the memory.

src/tool/hpcrun/main.c
```diff
@@ -671,8 +671,9 @@ hpcrun_fini_internal()
 
     TMSG(FINI, "process attempting sample shutdown");
 
-    SAMPLE_SOURCES(stop);
-    SAMPLE_SOURCES(shutdown);
+    //comment this dejan
```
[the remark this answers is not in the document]
dejangrubisic: We were discussing these when we were trying to solve that bug with memory corruption. I forgot to delete //.

src/tool/hpcrun/control-knob.c
dejangrubisic: Yes we have
  macro(**HPCRUN_CUDA_DEVICE_BUFFER_SIZE**, 8388608)     \
  macro(**HPCRUN_CUDA_DEVICE_SEMAPHORE_SIZE**, 65536)

if we don't use Nvidia, these macros won't be used, right, so why we need to hide them?

src/tool/hpcrun/sample-sources/nvidia.c
dejangrubisic: Ok, I turned them back. Should I delete them from knobs?

src/tool/hpcrun/gpu/gpu-trace-channel.h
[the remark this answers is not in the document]
Jokeren: You can declare `gpu_trace_channel_t` in `gpu-trace-channel.c` or `gpu-trace-channel-set.c`, using `typedef` to expose it, and implement interfaces to extract its members.

src/tool/hpcrun/control-knob.c
jmellorcrummey: They shouldn't appear as possible knobs unless we are using an nvidia device.

jmellorcrummey:
the control knobs still need work. I think that each sample source should register its control knobs with a control knob interface. That way, the control knobs available will be a function of the sample sources compiled in and available.

src/tool/hpcrun/control-knob.h
dejangrubisic: > the control knobs still need work. I think that each sample source should register its control knobs with a control knob interface. That way, the control knobs available will be a function of the sample sources compiled in and available.

Ok, should it be done similarly like registering device-finalizers? What parameters, connected to Sample Sources, should be there? Should it be something like sampling frequency?

jmellorcrummey: things like sampling frequency should not be in control knobs. the nvidia settings used by hpctoolkit (e.g. device buffer size, semaphore count) are examples of control knobs. for the gpu tracing, we could have a control knob (for instance) for number of helper threads.

dejangrubisic: Ok, I will try to implement that before our next meeting.

jmellorcrummey: it should be done similar to the device finalizers. each sample source can call register to register each knob. for nvidia sample source to register two knobs, it can just call register twice to register each of two knobs. the registry will hold a linked list of (name, value) pairs. the value can be represented by a void * in the registry. 

struct control_knob {
  char *name;
  void *(*value_parser)(char *value_string);
   char *(*value_printer)(void *value);
  void *value;
}

that way, a control knob can be anything: a string, an int, a double, or whatever. the void* can point to whatever storage is needed to hold the type.

dejangrubisic: Thanks, will do it!

PR #284, opened 2020-06-09
fix for Issue#280: Bug fix missing open cl call in thread root

aarontcopal2:
fix for Issue#280: Multithreaded version of OpenCL run does not display the last opencl operation at thread root.


aarontcopal2:
1st commit is made for code-review purposes. Code does not work, do not merge yet

PR #286, opened 2020-06-13
stub out gprof so that it doesn't interfere with hpcrun

jmellorcrummey:


jmellorcrummey:
@scottkwarren this patch eliminates the SEGV for the the picsarlite test on ppc64le. hpcrun data collection run now completes normally.

mwkrentel:
Funny, I can't get fib + gprof + hpcrun master to fail, not on ppc64le
(llnl.cs), not on x86_64, not REALTIME, not ITIMER, not segfault, not
deadlock, not even generate bad unwinds.

So, I'm not really sure what to do.  I can't tell if I'm missing
something or if you have something misconfigured.  That's part of why
I want to understand the mechanism of how it fails.

The current patch breaks the static case with multiple definitions of
__monstartup() and _mcleanup(), when compiled with -pg.

There are two ways to fix that.  One is to use the __wrap names like I
did with select and poll.  (You also need to add the names to the wrap
list in hpclink.)

The other is to have the hpclink script scan for the -pg option and
issue a warning or error, like we do with darshan.

jmellorcrummey:
> Funny, I can't get fib + gprof + hpcrun master to fail, not on ppc64le
> (llnl.cs), not on x86_64, not REALTIME, not ITIMER, not segfault, not
> deadlock, not even generate bad unwinds.

did you try https://github.com/hpctoolkit-tests/applications/fib-pg? This failed for me with master


> The current patch breaks the static case with multiple definitions of
> __monstartup() and _mcleanup(), when compiled with -pg.

I added a wrapper for _mcount as well and the stubs now support the static case as well

mwkrentel:
Ok, I can reproduce the problem with the fib-pg-bar test which uses
two files and a separate .so file.  It doesn't happen with the simple,
1-file fib.c.  So that says that there's something about a separate
function or a separate library that's an element of the conflict.

Also, it happens on ppc64le but not x86_64, and that's also currently
a mystery.

Anyway, the static case still doesn't work.  It needs to be --wrap to
be technically correct, anything else may (and should) fail.  This
seems to work or powerpc, but not on x86.  Not sure why, maybe weak
symbols or order on the link line.  On x86_64 I get:

```
$ ./hpclink  gcc  -g  -O  -static  -pg  fib.c  
hpclink: preliminary link 1 ...
hpclink: unable to link

/usr/lib/gcc/x86_64-redhat-linux/7/../../../../lib64/libc.a(_mcount.o):
In function `_mcount':
(.text+0x0): multiple definition of `_mcount'
```

It's quite easy, I could fix this for you, if you like.  :-)

mwkrentel:
Thinking about this a bit more, mcount() is too common a name to
unconditionally override.  Too easy for an app to use mcount() as a
function having no idea about gprof.  Maybe even the same for
_mcount().

If we really needed to override mcount() and _mcount(), it would have
to be an option for both hpcrun and hpclink.  I'd rather just tell
people, "don't use hpctoolkit and gprof at the same time."

jmellorcrummey:
We can use gotcha rather than LD_PRELOAD controlled by an option --ignore-pg

We can use the same option on hpclink

mwkrentel:
Alternate patch merged into develop in f6259b7d1eac on 2020-06-19.

Add option --disable-gprof to hpcrun and hpclink to override the gprof
functions and turn them into no-ops: __monstartup, _mcleanup, mcount
and _mcount.

PR #292, opened 2020-06-29
Handling extended numbering for ELF

mxz297:
Involved cases:

1. Extended section name string table index
2.Section index for each symbol

### DIFF
--- a/src/lib/binutils/RelocateCubin.cpp
+++ b/src/lib/binutils/RelocateCubin.cpp
@@ -80,6 +80,8 @@
 #include "ElfHelper.hpp"
 #include "RelocateCubin.hpp"
 
+#include "lib/prof-lean/elf-helper.h"
+
 
 
 //******************************************************************************
@@ -277,7 +279,7 @@ relocateLineMap
  Elf *elf,
  Elf_SectionVector *sections,
  Elf_SymbolVector *symbol_values,
- size_t section_string_index
+ elf_helper_t* eh
 )
 {
   GElf_Ehdr ehdr_v;
@@ -295,7 +297,7 @@ relocateLineMap
       GElf_Shdr shdr;
       if (!gelf_getshdr(scn, &shdr)) continue;
       if (shdr.sh_type == SHT_PROGBITS) {
-        const char *section_name = elf_strptr(elf, section_string_index, shdr.sh_name);
+        const char *section_name = elf_strptr(elf, eh->section_string_index, shdr.sh_name);
         if (strcmp(section_name, DEBUG_LINE_SECTION_NAME) == 0) {
           // remember the index of line map section. we need this index to find
           // the corresponding relocation section.
@@ -369,7 +371,7 @@ relocateDebugInfo
  Elf *elf,
  Elf_SectionVector *sections,
  Elf_SymbolVector *symbol_values,
- size_t section_string_index
+ elf_helper_t* eh
 )
 {
   GElf_Ehdr ehdr_v;
@@ -387,7 +389,7 @@ relocateDebugInfo
       GElf_Shdr shdr;
       if (!gelf_getshdr(scn, &shdr)) continue;
       if (shdr.sh_type == SHT_PROGBITS) {
-        const char *section_name = elf_strptr(elf, section_string_index, shdr.sh_name);
+        const char *section_name = elf_strptr(elf, eh->section_string_index, shdr.sh_name);
         if (strcmp(section_name, DEBUG_INFO_SECTION_NAME) == 0) {
           // remember the index of line map section. we need this index to find
           // the corresponding relocation section.
@@ -451,7 +453,7 @@ relocateSymbolsHelper
  GElf_Shdr *shdr,
  Elf_SectionVector *sections,
  Elf_Scn *scn,
- Elf_Scn *symtab_shndx_section
+ elf_helper_t *eh
 )
 {
   Elf_SymbolVector *symbol_values = NULL;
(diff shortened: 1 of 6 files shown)


src/tool/hpcrun/gpu/nvidia/cubin-symbols.c
```diff
@@ -221,12 +243,24 @@ relocateSymbols
   GElf_Ehdr ehdr_v;
   GElf_Ehdr *ehdr = gelf_getehdr(elf, &ehdr_v);
   if (ehdr) {
+    // If .symtab_shndx section exists, the section index of a symbol
```
jmellorcrummey, line 246: I think that it would be better to have a class that hides whether there is a SHT_SYMTAB_SHNDX section internally.

The class constructor could include this loop to initialize its symtab_shndx_section field.

src/tool/hpcrun/gpu/nvidia/cubin-symbols.c
```diff
@@ -221,12 +243,24 @@ relocateSymbols
...
+        break;
+      }
+    }
     for (int i = 0; i < sections->nsections; i++) {
       Elf_Scn *scn = sections->sections[i];
       GElf_Shdr shdr;
       if (!gelf_getshdr(scn, &shdr)) continue;
       if (shdr.sh_type == SHT_SYMTAB) {
-	      symbol_values = relocateSymbolsHelper(elf, ehdr, &shdr, sections, scn);
+	      symbol_values = relocateSymbolsHelper(elf, ehdr, &shdr, sections, scn, symtab_shndx_section);
```
jmellorcrummey, line 263: that class would be passed to relocateSymbolsHelper.

src/tool/hpcrun/gpu/nvidia/cubin-symbols.c
```diff
@@ -187,19 +188,40 @@ relocateSymbolsHelper
...
+    symtab_shndx_data = elf_getdata(symtab_shndx_section, NULL);
+  }
+
   Elf_Data *datap = elf_getdata(scn, NULL);
   if (datap) {
     symbol_values = newSymbolsVector(nsymbols);
     for (int i = 0; i < nsymbols; i++) {
       GElf_Sym sym;
-      GElf_Sym *symp = gelf_getsym(datap, i, &sym);
+      Elf32_Word xndx;
```
jmellorcrummey, line 204: The lookup method in the class would take (datap, i, &sym, unsigned long &index) and return symp.

src/tool/hpcrun/gpu/nvidia/cubin-symbols.c
```diff
@@ -187,19 +188,40 @@ relocateSymbolsHelper
   } else {
     return NULL;
   }
+
```
jmellorcrummey, line 191: this would also be in the constructor of the class.

src/lib/binutils/RelocateCubin.cpp
```diff
@@ -535,23 +563,37 @@ static Elf_SymbolVector *
...
-  Elf_SectionVector *sections
+  Elf_SectionVector *sections,
+  size_t section_string_index
 )
 {
   Elf_SymbolVector *symbol_values = NULL;
   GElf_Ehdr ehdr_v;
   GElf_Ehdr *ehdr = gelf_getehdr(elf, &ehdr_v);
   if (ehdr) {
+    // If .symtab_shndx section exists, the section index of a symbol
```
jmellorcrummey, line 574: this would be in the constructor of the class passed in.

src/lib/binutils/RelocateCubin.cpp
```diff
@@ -459,34 +464,56 @@ relocateSymbolsHelper
...
+
   if (datap) {
     symbol_values = new Elf_SymbolVector(nsymbols);
-
+    symbol_section_index = new Elf_SymbolSectionIndexVector(nsymbols);
     // Update symbol offsets
     for (int i = 0; i < nsymbols; i++) {
       GElf_Sym sym;
-      GElf_Sym *symp = gelf_getsym(datap, i, &sym);
+      Elf32_Word xndx;
```
jmellorcrummey, line 481: this is handled by the lookup method in the class.

src/lib/binutils/RelocateCubin.cpp
```diff
@@ -505,15 +532,15 @@ relocateSymbolsHelper
         if (symtype != STT_FUNC || symbind != STB_GLOBAL) {
           continue;
         }
-        int64_t s_offset = sectionOffset(sections, section_index(sym.st_shndx));
+        int64_t s_offset = sectionOffset(sections, section_index((*symbol_section_index)[i]));
```
jmellorcrummey, line 535: here, we would just pass "section_index" rather than indexing in the vector.

src/lib/binutils/RelocateCubin.cpp
```diff
@@ -505,15 +532,15 @@ relocateSymbolsHelper
...
         if (s_offset != 0) {
           // Sort functions in the same section
           std::vector<int64_t> offsets;
           for (int j = 0; j < nsymbols; ++j) {
             GElf_Sym nsym;
             GElf_Sym *nsymp = gelf_getsym(datap, j, &nsym);
             if (nsymp) {
-              if (GELF_ST_TYPE(nsym.st_info) == STT_FUNC && nsym.st_shndx == sym.st_shndx) {
+              if (GELF_ST_TYPE(nsym.st_info) == STT_FUNC && (*symbol_section_index)[i] == (*symbol_section_index)[j]) {
```
jmellorcrummey, line 543: support a symbolIndexToSymbolSection(i) method in the class.

mxz297:
> I think that we could just have an elf abstraction class and pass only that around. It would be easy to make all of the elf inquiries methods in this class. Then, we don't have to have all of the details of extended section counts, presence of a SHT_SYMTAB_SHNDX section, and the difference in the string extraction all exposed, and duplicated in multiple places.

I am good with doing more abstraction to hide details about ELF, but if we plan to do so, I would prefer to write it in C so that the hpcrun code can reuse the same abstraction. I don't want to program C++ abstraction that will later be replaced with C.

jmellorcrummey:
writing the abstraction in C would be fine. it would’ve been better if elfutils didn’t expose these details.

mxz297:
@jmellorcrummey Pushed a commit that abstracts the extended numbering handling in elf_helper functions.

src/lib/prof-lean/elf-helper.h
```diff
@@ -0,0 +1,91 @@
...
+
+#ifndef ELF_HELPER_H
+#define ELF_HELPER_H
+
+#ifdef __cplusplus
+extern "C" {
+#endif
+
+
+typedef struct elf_helper {
```
jmellorcrummey, line 61: it would be better to move the structure definition into the .c file, leaving just

`typedef struct elf_helper_t elf_helper_t`

in the header.

src/tool/hpcrun/gpu/nvidia/cubin-symbols.c
```diff
@@ -280,12 +287,14 @@ computeCubinFunctionOffsets
   Elf_SymbolVector *symbols = NULL;
   elf_version(EV_CURRENT);
   Elf *elf = elf_memory(cubin_ptr, cubin_len);
+  elf_helper_t eh;
```
mxz297, line 290: @jmellorcrummey I find that there is a good reason to keep the `elf_helper_t` definition in the header file. If we move the definition to the .c file, I will need an interface allocation function to return a pointer to `elf_helper_t`. Here we are in hpcrun. So we don't want to use malloc if possible in the allocation function. And since the code will be used by hpcstruct, we cannot use hpcrun_malloc in the allocation function. 

The strategy I am using is to put the type definition in the header file, declare it on stack, passed the pointer of the stack variable around.

jmellorcrummey: good point. I agree.

PR #304, opened 2020-07-30
Master gpu trace

dejangrubisic:
After fixing majority of identation issues, this is the final version for merging to develop branch.


src/tool/hpcrun/gpu/gpu-trace-channel.c
```diff
@@ -140,13 +143,27 @@ gpu_trace_channel_signal_consumer_when_full
...
 
+struct thread_data_t *
+gpu_trace_channel_get_td(gpu_trace_channel_t *ch)
+{
+  return ch->td;
+}
+
+
+void
+gpu_trace_channel_set_td(gpu_trace_channel_t *ch, thread_data_t *td)
```
jmellorcrummey, line 154: this is a bad name since there is already a gpu_trace_channel_set. I think the td should get added to the trace channel when it is created rather than putting it in afterward. should it be passed to the constructor, or have the td acquired within the constructor?

src/tool/hpcrun/gpu/gpu-trace.c
```diff
@@ -332,52 +334,54 @@ gpu_trace_stream_id
...
 
-static thread_data_t *
+thread_data_t *
 gpu_trace_stream_acquire
 (
  void
 )
 {
-  thread_data_t* td = NULL;
+  thread_data_t *td_self = hpcrun_get_thread_data();
```
jmellorcrummey, line 350: there seem to be two td instances in play here: td and td_self. why?

src/tool/hpcrun/sample-sources/nvidia.c
```diff
@@ -375,19 +371,12 @@ METHOD_FN(process_event_list, int lush_metrics)
 			    &device_finalizer_shutdown);
 
   // Get control knobs
-  int device_buffer_size =
-    control_knob_value_get_int(HPCRUN_CUDA_DEVICE_BUFFER_SIZE);
+  int device_buffer_size;
+  assert(control_knob_value_get_int("HPCRUN_CUDA_DEVICE_BUFFER_SIZE", &device_buffer_size) == 0);
```
jmellorcrummey, line 375: putting this in assertions is a bad idea. assertions can get elided if compiled with NDEBUG defined.

src/tool/hpcrun/sample-sources/cuda.c
```diff
@@ -336,7 +336,7 @@ METHOD_FN(process_event_list, int lush_metrics)
     PAPI_event_code_to_name(self->evl.events[i].event, buffer);
     TMSG(CUDA, "metric for event %d = %s", i, buffer);
     int metric_id = /* weight */
-      hpcrun_set_new_metric_info_and_period(cuda_kind, strdup(buffer),
+      hpcrun_set_new_metric_info_and_period(cuda_kind, strdup(buffer),						//dejan: this is setting metric into hpcviewer column
```
jmellorcrummey, line 339: the comment should have your name elided. isn't this sample source dead code? I don't think this belongs in this commit.

src/tool/hpcrun/gpu/gpu-trace-channel.c
```diff
@@ -68,6 +68,8 @@
 #include "gpu-trace.h"
 #include "gpu-trace-channel.h"
 #include "gpu-trace-item.h"
+#include "gpu-print.h"
+#include "thread_data.h"
```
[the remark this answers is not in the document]
dejangrubisic: There is one PRINT statement in the function **gpu_trace_channel_consume**, that is defined in gpu-print.h. Should I change it to something else?

src/tool/hpcrun/gpu/gpu-trace-channel.c
dejangrubisic: I am able to move gpu_trace_channel_set to:
gpu_trace_create -> gpu_trace_alloc -> gpu_trace_channel_alloc:

> channel->td = gpu_trace_stream_acquire(); <<< this is basically what gpu_trace_channel_set is doing so I can delete it.

However, I still need a getter function for getting the td out of channel. Is this name ok:
gpu_trace_channel_get_td () ?

src/tool/hpcrun/gpu/gpu-trace.c
dejangrubisic: Keren said that this part of the code, we use a helper thread to set thread-local data for every other thread. Inside **hpcrun_threadMgr_data_get** which is few lines below, we use pthread_setspecific so I believe that without saving td in td_self, we would overwrite it.

src/tool/hpcrun/sample-sources/cuda.c
dejangrubisic: I am not sure how that ended up in this commit but I deleted it.

src/tool/hpcrun/gpu/gpu-trace.c
jmellorcrummey: I understand now. You want to avoid losing the thread data of the helper thread. Rather than than coping with the fact that 

 hpcrun_threadMgr_data_get(id, NULL, &td, true);

has a side effect due to a pthread_setspecific and saving/restoring before/after, wouldn't it be better to have a variant of this routine that doesn't perform the side effect? That might mean 

- one routine that never performs this assignment and its caller is responsible for the side effect when wanted
- two routines - one that does everything except the side effect, and the second, which is a wrapper around the first additionally performs the side effect
- one routine that conditionally performs the side effect based on a boolean argument passed to it

Any of those options sound better than saving/restoring to cope with the side effect.  If the saving/restoring is indeed the best option, it deserves a few lines of comments  in this routine noting that the save/restore is to cope with the side effect.

dejangrubisic: I will check then with Keren and come back to you once we find best soluiton.

Jokeren: We can check with @laksono to see if change `hpcrun_threadMgr_data_get` is the best option. If not, I agree that we can add a more comprehensive comment.

laksono: My preference is the third option: one routine (`hpcrun_threadMgr_data_get`) that conditionally performs the side effect based on an `int option_policy` argument passed to it.
The reason I prefer `int option_policy` is just in case in the future we want to handle another case, and we don't want to add another boolean argument.

dejangrubisic: Do I need to change every call to hpcrun_threadMgr_data_get to wraper function with argument(preserve_self_td=false) ?

dejangrubisic: Is it hpcrun_threadMgr_data_get_safe a good name for wraper?

PR #314, opened 2020-08-25
Use and prefer versioned Python interpreters for stringify.py.

blue42u:
Needed to build on recent Debian, which has removed `python` in favor of the versioned dichotomy `python2` and `python3`.

Also fixes an issue where the build would fail strangely in compilation if Python wasn't available on the first `make` invocation.

### DIFF
--- a/src/tool/hpcstruct/Makefile.am
+++ b/src/tool/hpcstruct/Makefile.am
@@ -176,8 +176,10 @@ MOSTLYCLEANFILES = $(MYCLEAN)
 # Other rules
 #############################################################################
 
+PYTHON ?= $(shell which python3 || which python2 || which python)
+
 %.h : $(srcdir)/%.txt
-	$(srcdir)/stringify.py < $<  > $@
+	$(PYTHON) $(srcdir)/stringify.py $< $@
 
 
 #############################################################################
--- a/src/tool/hpcstruct/Makefile.in
+++ b/src/tool/hpcstruct/Makefile.in
@@ -1100,8 +1100,10 @@ endef
 # Other rules
 #############################################################################
 
+PYTHON ?= $(shell which python3 || which python2 || which python)
+
 %.h : $(srcdir)/%.txt
-	$(srcdir)/stringify.py < $<  > $@
+	$(PYTHON) $(srcdir)/stringify.py $< $@
 
 %.cpp.pp : %.cpp
 	$(CXXCPP) $(MYCPPFLAGS_0_CXX) $< > $@
--- a/src/tool/hpcstruct/stringify.py
+++ b/src/tool/hpcstruct/stringify.py
@@ -4,5 +4,7 @@ import sys
 import re
 
 end = re.compile('\n')
-for line in sys.stdin:
-    print('"' + end.sub(r'\\n"', line))
+with open(sys.argv[1], 'r') as infile:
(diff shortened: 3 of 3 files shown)


jmellorcrummey:
@blue42u If we are being pedantic about python, it would seem that python should be a spack dependence and that we should use the python specified as a spack dependence. 

What do both @blue42u and @mwkrentel  think?

jmellorcrummey:
While we are adjusting the use of python here, perhaps it would be better to generate the .h files in the build directory rather than the source directory. I put them in the source directory because of my ignorance how to bully the build system correctly.

blue42u:
@jmellorcrummey  The .h files appear in the build directory already, that doesn't appear to be an issue.

I'm ambivalent on the Spack issue, Spack itself requires a Python and we pull a Python through `intel-xed` anyway. A script this short doesn't care which Python it uses, and it only needs it during the build.

mwkrentel:
The classic autotools solution would be to search for a suitable
python in top-level configure and then pass that into the Makefiles.
But I'm ok with @blue42u's solution.  That's what spack does.

Or, just write it in perl.

mwkrentel:
> Also fixes an issue where the build would fail strangely in
> compilation if Python wasn't available on the first make invocation.

@blue42u  Huh?  How is this possible?
Is this in a spack or autotools build?

Spack heavily manipulates your environment and path during a build, so
maybe, but I don't see how this is possible in a configure + make +
make install build.

blue42u:
@mwkrentel This is in a manual autotools build using `--with-spack`. So if you don't have `python` on the PATH, the first `make` will fail from that, but the files won't be regenerated on the second `make` (since the redirection creates files), so it'll go on its merry way until a compilation syntax error.

Also, Perl? It'd be easier to swap it out for a shellscript. If you use Autotools you have a POSIX shell available.

mwkrentel:
Ok, I understand.  It will fail, just fail strangely.  From your
original comment, it sounded like make would fail (the "first" time),
but you restart make and then it would work.  That would suggest a
dependency issue.

mwkrentel:
@blue42u  @jmellorcrummey 
I guess no one actually tested this with python2.  (I didn't.)  :-(

The syntax here is illegal syntax with python 2.x.  This is on po
where /usr/bin/python is 2.7.5 and there is no python3 (not without a
module).

```
which: no python3 in (/opt/apps/software/Core/GCCcore/5.4.0/bin:/home/krentel/cmake/bin:/home/krentel/viewers/install/bin:/home/krentel/autotools/bin:/home/krentel/Packages/autotools/bin:/opt/apps/software/Core/xalt/default/bin:/usr/local/bin:/usr/bin:/usr/local/sbin:/usr/sbin:/opt/ibutils/bin:/home/krentel/bin)

/usr/bin/python2 ../../../../src/tool/hpcstruct/stringify.py ../../../../src/tool/hpcstruct/usage.txt usage.h
  File "../../../../src/tool/hpcstruct/stringify.py", line 10
    print('"' + end.sub(r'\\n"', line), file=outfile)
                                            ^
SyntaxError: invalid syntax
  File "../../../../src/tool/hpcstruct/stringify.py", line 10
    print('"' + end.sub(r'\\n"', line), file=outfile)
                                            ^
```

I'll just take python out of the equation and write it in shell.

PR #317, opened 2020-09-04
Step one of hiding hpcrun's functions and libraries.  This much works

mwkrentel:
around the deadlock with openmpi 4.x and UCX on stria at SNL.

  1. configure.ac tests if libunwind and lzma have a static (.a)
  archive compiled with -fPIC.  If yes, then proceed, else issue a
  warning and build libhpcrun.so as we previously do.

  2. Copy libunwind.a and lzma.a into src/extern and use objcopy to
  --redefine-sym mmap() and munmap() into hpcrun_real_mmap and
  hpcrun_real_munmap.  Using a separate directory is an easy way to
  make sure that these are built before hpcrun.

  3. In hpcrun/extern-real, write hpcrun_real_* functions to call the
  real versions via syscall().

  4. Edit hpcrun/Makefile.am, hpcrun, hpclink to allow both old and
  new versions of linking.

Remove the --enable-libunwind-primary option and clean up the related
hash-defines.  Libunwind is now too intertwined in the rest of the
code to be optional.

Not yet done: localize the symbols in hpcrun, and stripping the line
map without stripping the EH frames.

----------

ping  @jmellorcrummey 
I rebased and squashed my patches onto the latest version of master.
So, unless someone else comes along with a commit to master, it should
merge as a fast-forward merge with linear history.


mwkrentel:
The diffs are large because I touched the build system and created
three new Makefiles.  The only files that I actually changed are the
following, the rest are all auto-generated.  And most of changes are
small, except for configure.ac and maybe hpcrun/Makefile.am.

```
Makefile.am
configure.ac
src/Makefile.am
src/extern/Makefile.am
src/extern/libunwind/Makefile.am
src/extern/lzma/Makefile.am
src/lib/support-lean/Makefile.am
src/tool/hpcrun/Makefile.am
src/tool/hpcrun/extern-real/mmap.c
src/tool/hpcrun/main.c
src/tool/hpcrun/scripts/hpclink.in
src/tool/hpcrun/scripts/hpcrun.in
```

jmellorcrummey (approved):
I found the setting LIBUNWIND_COPY=no confusing. I had to go read your whole configure file to find out what it meant. It means copy libunwind.so into our install directory. With the fPIC .a file, you copy that and change some some of the names it uses. Apparently, that copy doesn't count.

I think a name like LIBUNWIND_COPY_SHARED_LIBRARY would be better.

For the MBEDTLS library, it seems that we are just using the pic .a file and never copy and install its shared library as well?

What external dependences on shared libraries does that leave us with at runtime?

jmellorcrummey (commented):
will this work without libunwind on ppc64? we don't use libunwind on ppc64.

mwkrentel:
> I found the setting LIBUNWIND_COPY=no confusing. 

Ok, fair enough.  What I'll do is move LIBUNWIND_COPY to the end of
libunwind, after we've tested everything else.  So, we'll copy
libunwind.so if and only if libunwind exists but fails fpic.  That's
cleaner that turning it on and then back off.

Note: the copy variables are used for the top-level Makefile.am which
is where we do the copy.  That's the place to look.

As for mbedtls, when we added it initially, we required it to be
compiled with -fPIC.  It's always been like that for mbedtls,
it's -fPIC or else fail.

For libunwind and lzma, I rewrote configure so that it would test for
and accept -fPIC or not and act accordingly.  That's smoother and more
robust than a hard transition that always requires it.

mwkrentel:
> will this work without libunwind on ppc64? 
> we don't use libunwind on ppc64.

Well, you might naively think so.  :-)

With master, if you configure with --without-libunwind, the build
fails with:

```
In file included from ../../../../src/tool/hpcrun/unwind/ppc64/unw-datatypes-specific.h:50:0,
                 from ../../../../src/tool/hpcrun/unwind/common/unw-datatypes.h:51,
                 from ../../../../src/tool/hpcrun/unwind/common/backtrace.h:73,
                 from ../../../../src/tool/hpcrun/cct/cct.h:84,
                 from ../../../../src/tool/hpcrun/epoch.h:54,
                 from ../../../../src/tool/hpcrun/thread_data.h:63,
                 from ../../../../src/tool/hpcrun/safe-sampling.h:104,
                 from ../../../../src/tool/hpcrun/sample-sources/ga-overrides.c:78:
../../../../src/tool/hpcrun/unwind/common/std_unw_cursor.h:56:10: 
fatal error: libunwind.h: No such file or directory
 #include <libunwind.h>
```

I dug a little deeper.  In the .o files from our code in
hpcrun/unwind/common, these functions are referenced.

```
$ cd build/src/tool/hpcrun/unwind/common
$ nm lib*.o | grep  -e ' U ' | sort 
   U _ULppc64_apply_reg_state
   U _ULppc64_apply_reg_state
   U _ULppc64_get_reg
   U _ULppc64_get_reg
   U _ULppc64_set_reg
   U _ULppc64_set_reg
   U _ULppc64_step
   U _ULppc64_step
   ...
```

So, hpcrun/unwind/common is pulling in the libunwind code.  It must be
code that we reference but never run.

I tried taking out the libunwind libraries from hpcrun/Makefile.am.
The dynamic case seems to run, but hpclink fails with undefined
references to the above functions.  This is consistent with code that
we refer to but never run.

All the above is with master, it's what we do now.  What does my patch
do?  Exactly what we do now in master.  It preserves the status quo,
not any better, not any worse.

So, we could try extracting libunwind from the powerpc case, but it's
not as simple as "turn it off."  And it's completely separate from the
current PR.

I remember maybe 3-4 years ago, we did turn off libunwind for powerpc
(due to problems on blue gene), but I couldn't find the commit.
Although we don't run the code, it's still intertwinded with the code
base.

jmellorcrummey:
ok. proceed with the pull request. we will disentangle libunwind from hpctoolkit on ppc at another time. we’ll just record that as software debt.

mwkrentel:
@jmellorcrummey I updated LIBUNWIND_COPY.
Are we ok to merge now?

PR #318, opened 2020-09-08
AMD GPU support with rocm 3.7

mxz297:
1. Update the use of roctracer to reflect API changes from rocm 2.x
2. Use AMD debug library to query binary URI and extract AMD GPU binary
3. Ignore monitoring threads created by AMD libraries

### DIFF
--- a/src/tool/hpcrun/module-ignore-map.c
+++ b/src/tool/hpcrun/module-ignore-map.c
@@ -102,7 +102,7 @@
 #define PRINT(...)
 #endif
 
-#define NUM_FNS 3
+#define NUM_FNS 6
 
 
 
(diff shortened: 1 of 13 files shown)


mxz297:
@jmellorcrummey To refactor module ignore map to be filled in dynamically depending on the command line argument, we need to reorder things in the hpcrun initialization step. I believe Dejan's PAPI work requires some reordering in hpcrun init and he has done some of the reording. I would prefer deferring the refactor of module ignore map after merging Dejan's PAPI's branch to avoid conflicts.

src/tool/hpcrun/module-ignore-map.c
```diff
@@ -102,7 +102,7 @@
 #define PRINT(...)
 #endif
 
-#define NUM_FNS 3
+#define NUM_FNS 6
```
jmellorcrummey, line 105: add a FIXME note that indicates we will later refactor this code to make the NUM_FNS a variable and the table of functions to be looked up can be a linked list, where any GPU can indicate that its functions should be added to the module ignore map when that type of GPU is being monitored.

jmellorcrummey (changes requested):
Overall, everything looks good.

I think this is good enough to have as a prototype in develop, though it needs some work before it makes it to master. At present, my suggestion is to add a collection of FIXME comments where indicated so that we don't forget the refactoring needed before it is ready for prime time.

jmellorcrummey (approved):

jmellorcrummey:
I am fine with it at present. Do you want to squash and merge or keep the commits separate. If you squash and merge, you might want to drop the comment about the reverse engineering, since I think that code has been replaced by the debug-api.

mxz297:
Squash and merge is better for this PR. But I am not sure how I can rewrite the commit message of the "squash and merge" commit. I think it is automatically generated as a listing of all commits in the PR.

jmellorcrummey:
squash and merge done. comments related to internal commits that are not important descriptors of the final result (e.g. fix compilation, reverse engineering which was done and then replaced) have been deleted.

PR #320, opened 2020-09-15
Build with Meson

blue42u:
This is the ~~highly anticipated~~ build system conversion, which adds an alternative to the usual Autotools-based system using Meson. This is a nearly purely additive patch to prevent potential issues during merges, the original Autotools' functionality is unmodified.

There are a number of reasons to move away from Autotools:
 - Modern build systems(' languages) are more terse and powerful. This patch converts over 5600 (non-comment) lines of Autotools source into just over 1900 (non-comment) lines of Meson and related support code. Less code means easier maintainability and less chances for trivial errors.
 - Modern build systems are acutely aware of their sources, both by regenerating automatically upon edits to such files and ensuring no build outputs end up in the source repository. Autotools not only requires an `autoreconf` on build system edits, but also uses `make` which has no concept of "cyclic dependency errors." Purging build outputs from the repository ensures all contained files are "human" in nature, allowing for more readable patches and less opportunities for mistakes (such as the minor debacle during the NVIDIA support merges).
 - **ADD:** Ninja (modern `make` replacement) is very careful about ensuring the build directory is consistent at all times: it caches build commands and will regenerate files who's commands change, and it caches output files and removes files which have disappeared between runs. Autotools is stuck with `make` which does neither of these, nor does it add its own support to do so like CMake does. Having to ask "did you `make clean`/nuke your build dir?" every time someone reports a really weird bug gets tiring after a while.
 - Using a build system gives access to the suite of features supported by that build system. Modern build systems have more abstractions and thus are more robust against minor diversity issues between systems (including cross compiling). In contrast, Autotools forces many extensions and paths to be specified manually, and real dependency management is just plain difficult (not everyone ships a `pkg-config` file).

And in case that wasn't convincing enough, I quote the "rationale" given when [the illustrious systemd made the jump three years ago](https://github.com/systemd/systemd/commit/5c23128daba7236a6080383b2a5649033cfef85c):
> It's crucial that we can build [HPCToolkit] using Visual Studio 2010!
>
> ... er, wait, no, that's not the official reason. We need to shed old systems
> by requring python 3! Oh, no, it's something else. Maybe we need to throw out
> 345 years of knowlege accumulated in autotools? Whatever, this new thing is
> cool and shiny, let's use it.

---

As for why Meson and not the more (in)famous CMake, Meson supports a more complete language (with first-class objects!), abstracts away many easily problematic functionalities (e.g. user-defined cache entries), and generally uses far less conditionals to perform the same tasks. These make it much easier to be confident that the build system will perform as expected (or at least predictably) in unknown environments. CMake also doesn't natively support using both the static and shared library for a dependency in the same build, working around which turns out to be very complicated and fragile to perform correctly.

---

> This is not complete, I'm throwing it out here for your amusement and critique.

In addition to the (likely) bugs there are a few major differences:
 - `lib/hpctoolkit/ext-libs` is not implemented, Meson is very adamant about not giving real paths for external libraries. The binaries are stitched with `RUNPATH`s automagically instead, and applicable launch scripts are replaced by their raw executables. (Exactly how this affects libdwarf handling is not fully tested yet.)
 - Meson clears `RUNPATH`s as part of installation, which means an install tree will not function without first loading all the dependencies into `LD_LIBRARY_PATH`. For convenience usable binaries and launch scripts are copied to the `bin/` subdirectory of the build tree. (Spack patches this part of Meson out, so Spack installs work and installs with a Spack Meson might.)
 - For the above reasons, the `hpcrun` launch script has many of its path bits removed. It also uses `ldd` to find `libmonitor` for the explicit `LD_PRELOAD`. To keep the patch additive the modified script source is saved in the sibling `hpcrun.meson.in`.
 - Also for the above reasons, the mechanism behind the `hpclink` script had to be heavily rennovated. The modified script uses `pkg-config` to generate all of the link arguments, `.pc` files for which are generated semi-automatically by Meson. The modified script source is saved in `hpclink.meson.in`.
 - **ADD:** The change in 2b99620266978bf16b3593aec46586a6ac9a3957 to hide most of the internal symbols cannot be ported directly to Meson, and for now this change is skipped in this PR. For the shared case the preferred option is to use in-source attributes to specify the desired visibility, for the static case there is no solution. A proof-of-concept patch is given in 895bf88d27f8e7763af72bda0b00d56845a6bf9c, which generates nearly the same symbol listing as the current Autotools (minus symbols from prof-lean etc.).
 - **ADD:** The change in 764b45dfc33eade51f0f6a6a3000e8915b403084 to inject a custom path for memkind is currently not ported for the same reasons as `ext-libs`.
 - **ADD:** ~~The side-effects for 9b4801f239badeaa0b47d701f1414f979715ac39 and related commits on the Meson build system have not been applied yet, as such using `hpcstruct` on a measurements directory may not work as intended. WIP.~~ The changes made by 9b4801f239badeaa0b47d701f1414f979715ac39 and related have now been ported to Meson, with non-additive changes to remove dependence on `HPCTOOLKIT_INSTALL_PREFIX`. These changes should be harmless even in a non-Meson environment.
 - **ADD:** The framework for standalone testing (testing outside of a build) is completely restructured for Meson. fbbb0b045c0ac839ce6e7f6033aca43d24bfc879 copied a separate manually-written Makefile to the standalone testing system (redundantly giving compile commands for the test binaries), this branch instead copies the whole directory and uses it as it's own Meson project. This removes the redundancy and allows use of Meson's testing utilities even in a `spack test` environment.

Otherwise the list of installed files matches closely to those from Autotools, excluding `.la` files, a rename for the static case (`libhpcrun.a -> libhpcrun_static.a`) and the new framework files for `hpclink`.

There are probably a number of missing conditionals, I mostly focused on the case where most dependencies are available and are the latest versions. In particular, Dyninst using libdwarf probably won't work.

---

Building with Meson is fairly simple, a quick-and-dirty build can be achieved with
```console
$ <spack> load --only package meson ninja cmake pkgconf  # if needed
$ meson build/ -Dspack=<spack>
$ ninja -C build/
```
where `<spack>` is the path to your Spack command (not the install root), or just `spack` if its on your PATH.

If you require more control over which packages are loaded and from where, instead use [Meson's "native" configuration files](https://mesonbuild.com/Native-environments.html). Create a file `my_native.ini` with contents such as:
```ini
[properties]
spack = 'spack'
# Everything should use the GCC 9.2 version
default_spec = '%gcc@9.2'
# Except libunwind and TBB, which should be GCC 10 and use PIC
libunwind_spec = '+pic %gcc@10'
intel-tbb_spec = '%gcc@10'
```
and then replace the above commands with:
```console
$ meson build/ --native-file my_native.ini
$ ninja -C build/
```

For even more control, give the full paths to the individual dependencies (not recommended for Spack packages):
```ini
# Use the libbfd installed here
libbfd_root = '/my/binutils'
# If its not installed in lib/ and include/, overload the search suffixes:
library_path_suffixes = ['lib64', 'lib', 'lib/x86_64-linux-gnu']
include_path_suffixes = ['include', 'include/x86_64-linux-gnu']
# Or alternatively specify the directories explicitly:
libbfd_librarydir = '/my/binutils/lib42'
libbfd_includedir = '/my/binutils/include'
```
Or, alternatively disable a dependency completely with:
```ini
libelf_enabled = false
```
NOTE: Arguments such as `--native-file` only take effect on the first invocation of Meson, changing the set of native files requires the build directory to be entirely removed first. On the other hand, Meson will remember the paths to such files and regenerate if they change, so creating an empty one for future adjustments isn't a bad idea. See the appropriate documentation for more details.

Currently the exception to the above rules is CUDA/CUPTI, which requires:
```
$ CUDA_HOME=/my/cuda meson build/ --native-file my_native.ini
```
to set the path to the CUDA installation. If you need to modify the path in an existing build directory, use
```
$ CUDA_HOME=/my/new/cuda meson build/ --wipe
```
Setting `cuda_enabled` to false as above will disable CUDA as one would expect.

---

For experimentation purposes, [my personal Spack fork](https://github.com/blue42u/spack/tree/hpctoolkit-meson) has a branch with a modified `hpctoolkit` package supporting only `@master` and building with Meson (i.e. this PR branch). In summary:
```console
$ git clone https://github.com/blue42u/spack.git -b hpctoolkit-meson spack-ht-meson
$ ./spack-ht-meson/bin/spack install hpctoolkit
$ ./spack-ht-meson/bin/spack load hpctoolkit
```
In addition this branch contains a handful of hacks intended to work around Meson-related Spack bugs present in `develop`. While this makes it unsuitable for a Spack PR, it suffices for testing.

**EDIT:** Previously this was a sibling `hpctoolkit-meson` package, this has been changed to modifications to the `hpctoolkit` package to ease updates as the upstream `hpctoolkit` package is updated. When updating a clone of this branch please always use `git pull --rebase` or `git fetch origin; git reset --hard origin/hpctoolkit-meson`.

---

Many of the `--enable-*` options have been converted into Meson options, which can be specified on the first run for a build directory and modified later with:
```console
$ meson configure build/ -D<option>=<value> ...
```
Running the above with no `-D` arguments prints all available options with current values and short descriptions. Most options retain the same name as the Autotools variant, the full current conversion table is as follows:
| instead of... | use... | by default... or notes |
| --------------- | ------- | ------------ |
| `--enable-all-static` | (never enabled) | on TODO |
| `--enable-develop` | `-Dbuildtype=debug -Dstrip=false` | disabled (`-Dbuildtype=release`) |
| `--enable-strip-debug` | `-Dstrip=true` | enabled |
| `--enable-endian` | (Meson builtin) | determined automatically |
| `--enable-mpi-search` | (Meson builtin) | see [3] below |
| `--enable-bgq` | (never enabled) | |
| `--enable-force-papi` | (always enabled) | |
| `--enable-papi-c` | `-Dpapi-c=enabled` | enabled if available |
| `--enable-papi-c-cupti` | `-Dpapi-c-cupti=enabled` | enabled if available |
| `--enable-perf-events` | `-Dperf-events=enabled` | enabled if available |
| `--enable-kernel-blocking` | `-Dkernel-blocking=enabled` | enabled if available |
| `--enable-back-end` | `-Dhpcstruct=disabled -Dhpcprof=disabled` | see [1] below |
| `--enable-hpcrun` | `-Dhpcrun=enabled` | see [1] below |
| `--enable-hpcrun-static` | `-Dhpclink=enabled` | see [1] below |
| `--enable-hpcserver` | `-Dhpcserver=enabled` | disabled (see [1] below) |
| `--enable-openmp` | `-Dopenmp=enabled` | enabled if compiler supports |
| `--enable-mpi-wrapper` | (never enabled) | on TODO |
| `--enable-lush` | `-Dlush=true` | currently no-ops, on TODO |
| `--enable-lush-pthreads` | `-Dlush-pthreads=true` | disabled (`false`) |
| `--enable-xop` | `-Dxop=true` | enabled (`true`) |
| `--enable-data-centric-tracing` | `-Ddata-centric-tracing=true` | disabled (`false`) |
| `--enable-devtools` | `-Ddevtools=enabled` or `=auto` | disabled (see [1] below) |
| `--enable-valgrind-annotations` | `-Dvalgrind-annotations=enabled` or `=auto` | disabled |
| `MPICXX=...` | `-Dmpi=enabled` | enabled if available (see [2] below) |
| `--with-cuda=...` | `-Dcuda-monitoring=enabled` | enabled if available |

[1] By default all four supported applications (`hpcrun`, `hpclink`, `hpcstruct` and `hpcprof`) are set to `auto`, i.e. compile and install if all dependencies are available. `hpcfnbounds` and `hpcfnbounds2` currently always follow this and cannot be disabled.  The remaining applications (`dotgraph`, `hpcproftt`, `hpclump`, `hpctracedump` and `hpcserver`) are by default hard-disabled and must be explicitly enabled (or set to `auto` to compile when possible).
[2] This controls both `hpcprof-mpi` and `hpcserver-mpi`, in addition to their individual non-MPI settings. So neither `-Dhpcprof=auto -Dmpi=disabled` nor `-Dhpcprof=disabled -Dmpi=auto` will compile `hpcprof-mpi`, but `-Dhpcprof=auto -Dmpi=auto` can if possible (and `=enabled` always will).
[3] Meson has built-in support for MPI as a dependency, which searches for a wrapper from the environment ([implementation](https://github.com/mesonbuild/meson/blob/master/mesonbuild/dependencies/mpi.py)). If there's some system that doesn't work with this, it would be easier to upstream an adjustment rather than attempt a workaround.


blue42u:
[This repository has moved to GitLab, please continue this PR there.](https://gitlab.com/hpctoolkit/hpctoolkit/-/issues/320)

blue42u:
The previous link was incorrect, apologies. The corrected link is [here.](https://gitlab.com/hpctoolkit/hpctoolkit/-/merge_requests/320)

PR #324, opened 2020-09-19
added support for gpu tuples

dejangrubisic:
Added support for gpu tags


src/tool/hpcrun/write_data.c
```diff
@@ -175,7 +176,9 @@ lazy_open_data_file(core_profile_trace_data_t * cptd)
 
   int rank = hpcrun_get_rank();
 
-  compute_profile_name(rank, cptd);
+  if (cptd->id < 500){
```
Jokeren, line 179: Why there's still this limitation?

src/tool/hpcrun/gpu/gpu-trace.c
```diff
@@ -90,9 +90,16 @@
 // type declarations
 //******************************************************************************
 
+typedef struct gpu_tag {
+ int device_id;
+ int context_id;
+ int stream_id;
+}gpu_tag;
```
Jokeren, line 97: `gpu_tag_t` might be better

src/tool/hpcrun/write_data.c
dejangrubisic: In this version threads with id < 500 are cpus, we can put a different thing in if clause like cptd->id_tuples == NULL

src/tool/hpcrun/gpu/gpu-trace.h
```diff
@@ -96,14 +96,16 @@ gpu_trace_init
...
+ int context_id,
+ int stream_id
 );
 
 
 void *
 gpu_trace_record
 (
- gpu_trace_t *thread_args
+ struct gpu_trace_t *gpu_trace
```
Jokeren, line 108: Why cannot you say `gpu_trace_t *gpu_trace`?

src/tool/hpcrun/write_data.c
Jokeren: Yeah, I would prefer the latter or an alternative check here. `500` CPU threads is quite limited.

src/tool/hpcrun/gpu/gpu-trace.h
dejangrubisic: will change it

src/tool/hpcrun/gpu/gpu-trace.c
dejangrubisic: changed

jmellorcrummey (changes requested):
for the gpu tuples, we don’t want to report the bogus thread id (e.g. 500). internally, we should use 1M instead of 500 (as per keren’s point, we don’t want them to overlap with  cpu thread indices). opencl and level 0, i think the abstraction is a queue rather than a gpu context and a stream. we should check with @mxz297 and @aarontcopal2  about that. rather than naming everything context and stream all the way through the gpu support, we might just call them gpu_context1 and gpu_context2. for the queue abstraction, i don’t think there is a second level value. we should have the gpu runtime specify in an initialization call to the gpu tracing infrastructure what the IDTUPLE kind values are for gpu_context1 and gpu_context2. we can add a IDTUPLE_NOKIND type that we use internally to indicate one context id (for instance that gpu_context2) is to be ignored and not recorded in a profile.

Jokeren:
@jmellorcrummey opencl does have the concept of  `context`. Each command queue is associated with a `context`.

https://github.khronos.org/OpenCL-CLHPP/classcl_1_1_command_queue.html#adcb116a4e1a53417dad3d3850c33d42f

oneapi, as far as I understand, is a high level abstraction which could use opencl and cuda as a backend. Thus, I think it is for to create two levels of abstraction here. We can rename the term `stream` to `queue` if that makes more sense.

@mxz297 , @aarontcopal2 correct me if I am wrong here.

mxz297:
@jmellorcrummey @Jokeren 

I agree that we should think about a more general representation. For NVIDIA, there are `device`, `context`, and `stream`. I don't quite know what a `context` means compared to a `stream`, so I cannot relate it to concepts for other GPUS. 

For AMD GPU, or HIP more specifically, there are `device` and `stream`, but there does not seem to be a `context`. 

For Level 0 v1.0 (https://spec.oneapi.com/level-zero/latest/core/PROG.html#device), there are `device`, `context`, `CommandQueue`, and `CommandList`. Based on documentation, L0's context is a logical object used by the driver for managing all memory, command queues/lists, modules, synchronization objects.

I feel there are two questions here:

1. How the terminology differ for different platforms? In particular, does the same word mean the same thing for different GPUs? 

2. What information is useful for the users to optimize their code?

src/tool/hpcrun/write_data.c
dejangrubisic: I have just added is_id_tuple_empty() check

dejangrubisic:
I agree with all your points. We could easily expand id_tuples with diferent types of abstractions that represent a tag. the only question is should we do that in cupti/rocm/opencl/level0_activity_translate and how would we send that informatinon to the tracing monitoring thread that writes data into files?

Jokeren:
> * How the terminology differ for different platforms? In particular, does the same word mean the same thing for different GPUs?
> *

Usually context does not matter from a programmer's perspective. When you hold multiple contexts in a single process, the kernels that belong to each context cannot launch concurrently. The GPU has to switch between contexts to launch kernels. So usually we use a context per process.

jmellorcrummey (approved):
accepting these changes to continue refining them.

PR #326, opened 2020-09-21
CUPTI fork() Handling Hotfix

blue42u:
Also adds a big comment with a description of the issue.

@jmellorcrummey Edit the comments as necessary if I missed something.


jmellorcrummey:
this addresses the problem that we saw with deep wave, which forks a process that has active CUPTI measurement. handling stopping and restarting CUPTI measurements at fork needs to be improved. for that matter, sample source method organization for all sample sources need improvement.

PR #327, opened 2020-09-22
L0 refactor

mxz297:
This PR refactors Level0 support to work with Intel OneAPI Beta 9 or newer. Older versions of OneAPI is no longer supported. 

We also used function override to intercept Level0's public API as previous tracing support in Level0 is now marked as experimental and will be removed in the future.

I have rebased this PR upon the latest LD_AUDIT commits and tested it with DPC++ samples.

### DIFF
--- a/src/tool/hpcrun/gpu/level0/level0-command-list-context-map.c
+++ b/src/tool/hpcrun/gpu/level0/level0-command-list-context-map.c
@@ -67,6 +67,10 @@
 // local data
 //******************************************************************************
 
+// TODO:
+// Replace mutual exclusion with more efficient
+// data structures.
+
 static level0_handle_map_entry_t *commandlist_context_map_root = NULL;
 
 static level0_handle_map_entry_t *commandlist_context_free_list = NULL;
--- a/src/tool/hpcrun/gpu/level0/level0-command-list-context-map.c
+++ b/src/tool/hpcrun/gpu/level0/level0-command-list-context-map.c
@@ -41,83 +41,91 @@
 //
 // ******************************************************* EndRiceCopyright *
 
-
-#ifndef level0_commandlist_map_h
-#define level0_commandlist_map_h
-
 //*****************************************************************************
 // system includes
 //*****************************************************************************
 
-#include <stdint.h>
+#include <assert.h>
+#include <string.h>
 
 //*****************************************************************************
 // local includes
 //*****************************************************************************
 
+#include "level0-command-list-context-map.h"
+#include "lib/prof-lean/spinlock.h"
 
-#include <level_zero/ze_api.h>
-#include <level_zero/zet_api.h>
-#include "level0-handle-map.h"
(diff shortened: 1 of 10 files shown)


src/tool/hpcrun/sample-sources/level0.c
```diff
@@ -195,8 +196,11 @@ METHOD_FN(finalize_event_list)
     monitor_real_exit(-1);
   }
 #endif
-
   level0_init();
+
+  // Init records
+  gpu_trace_init();
```
jmellorcrummey, line 202: shouldn't this only be done if tracing is enabled? is that handled inside the gpu tracing code?

src/tool/hpcrun/gpu/level0/level0-command-list-context-map.c
```diff
@@ -0,0 +1,131 @@
...
+static level0_handle_map_entry_t *commandlist_context_free_list = NULL;
+
+static spinlock_t commandlist_context_map_lock = SPINLOCK_UNLOCKED;
+
+//*****************************************************************************
+// interface operations
+//*****************************************************************************
+
+ze_context_handle_t
+level0_commandlist_context_map_lookup
```
jmellorcrummey, line 81: having this use mutual exclusion seems OK for now. I think it  should be marked with a FIXME to reconsider mutual exclusion at a later time.

jmellorcrummey (changes requested):
Consider whether tracing initialization should always be invoked, even if tracing is turned off.

Mark the command list map with a FIXME to reconsider mutual exclusion at a later time.

src/tool/hpcrun/sample-sources/level0.c
mxz297: After a further look, I am not sure... 

`gpu_trace_init` is also always called for nvidia code:

https://github.com/HPCToolkit/hpctoolkit/blob/develop/src/tool/hpcrun/sample-sources/nvidia.c#L272

And the implementation of `gpu_trace_init` does not seem to examine the command line argument for tracing.

src/tool/hpcrun/gpu/level0/level0-command-list-context-map.c
mxz297: Added TODO comments.

src/tool/hpcrun/sample-sources/level0.c
jmellorcrummey: I read the code. It simply initializes variables used to coordinate any tracing that may occur. Performing this initialization is harmless. I approve.

jmellorcrummey:
Do you want this squashed and merged, or is there merit to keeping the commits separate so that the gotcha code can be revived if we later decide we want to use that wrapping strategy?

mxz297:
Squashing is fine as I realized that I forgot to check-in the gotcha implementation, which were removed anyway...

PR #328, opened 2020-09-23
Rtld-audit Fallback Patch

blue42u:
Allow disabling the LD_AUDIT-based infrastructure using --disable-rtld-audit

Using this flag allows for hpcrun to still function on systems where LD_AUDIT
support in Glibc is currently non-functional. There will be R*PATH issues when
this is used, similar to the state of hpcrun previous to all this.

Note that with both this and the previous LD_AUDIT commit, libmonitor should be
configured with --disable-dlfns, regardless of the --enable-rtld-audit setting.

### DIFF
--- a/Makefile.in
+++ b/Makefile.in
@@ -424,6 +424,7 @@ datarootdir = @datarootdir@
 docdir = @docdir@
 dvidir = @dvidir@
 exec_prefix = @exec_prefix@
+export_ld_audit = @export_ld_audit@
 hash_fcn = @hash_fcn@
 hash_value = @hash_value@
 host = @host@
--- a/configure
+++ b/configure
@@ -734,6 +734,9 @@ TBB_LIB_DIR
 TBB_PROXY_LIB
 TBB_LFLAGS
 TBB_IFLAGS
+OPT_ENABLE_RTLD_AUDIT_FALSE
+OPT_ENABLE_RTLD_AUDIT_TRUE
+export_ld_audit
 OPT_PAPI_CUPTI_FALSE
 OPT_PAPI_CUPTI_TRUE
 OPT_PAPI_COMPONENT_FALSE
@@ -1043,6 +1046,7 @@ enable_papi_c_cupti
 with_perfmon
 enable_perf_events
 enable_kernel_blocking
+enable_rtld_audit
 with_tbb
 with_xed
 with_xerces
@@ -1748,6 +1752,8 @@ Optional Features:
                           force enable or disable kernel blocking support in
                           perf events (normally 4.3 or later), only needed if
                           fails to auto-detect correctly
+  --enable-rtld-audit     enable or disable the use of LD_AUDIT for monitoring
+                          of loaded binaries
   --enable-back-end       build only those tools needed to run hpcrun and
                           hpclink on the back-end compute nodes
   --enable-hpcrun         build hpcrun (default yes)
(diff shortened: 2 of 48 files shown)


jmellorcrummey (approved):
Changes to enable external forcing of initialization were added.

Changes to dlopen and dlopen wrappers to have only a single call to the wrapped function were added.

This certainly needs some adjustment for the static case, but I think that Mark can address that in a second pass after accepting this code to the develop branch.

jmellorcrummey:
@mwkrentel @blue42u  we now have two regression tests that expose auditor failure

1. the simple ARM regression test I submitted to Ben
2. a test we coded up yesterday that uses a trivial auditor to monitor a main program that simply calls dlmopen twice on libm.so.6

Do either of our build systems (auto tools, meson) support an execution test at runtime that can be used to determine if these two audit bugs are present or not to help make a configuration decision?

blue42u:
I know how to do it in Meson, its a matter of "running" the link command to spin out the trivial auditor and then running the tests with the `LD_AUDIT` environment variable set. If it works without crashing, we can use the real auditor.

From what @mwkrentel has mentioned in the past, I assume Autotools has similar language. I just don't know the magic M4 for it.

mwkrentel:
For any build system, runtime tests are discouraged because at best
you're testing the build machine, not the run machine (compute nodes).
For a true cross compile, the runtime test won't even run.

AFAIK, bug (1) happens on ARM and only ARM, so you could just test the
host arch, allowing for a manual override, of course.

What is bug (2) and where does it happen?

But yes, autotools does allow runtime tests, they just discourage
using them.

blue42u:
Doesn't Autotools have a CROSS flag that's set when cross-compiling? I'm no expert (and can't seem to find it again), but I thought it was fairly simple to test that case.

Meson doesn't let you run compiled outputs if you're cross-compiling (unless the user has given a wrapper script to permit so), and also has a simple flag (`meson.is_cross_build()`) to check for that case.

More on topic, bug (2) is something discovered yesterday where `dlmopen` is generally broken on all but the latest Glibc when `LD_AUDIT` is active: https://sourceware.org/bugzilla/show_bug.cgi?id=26076. Aaron ran into it after merging `develop`, it turned out that `libgtpin` decided to use `dlmopen` as part of its init constructor, and so crashed because of the bug in Glibc.

Bug (1) doesn't occur on ARM + Ben Woodard's patch, so I'd prefer to do a proper test for it as well. That will also help the transition as machines have their Glibc updated.

mwkrentel:
Yes, you can ask autotools if you're cross compiling.  But all that
means is that you know enough not even to try a run test.

Even if you're not cross compiling, no build system will tell you the
precise features of the compute nodes.  Not without launching a job on
them.

jmellorcrummey:
@mwkrentel I would like to approve this commit because it will fix issues caused by transitioning the develop branch to LD_AUDIT, which exposed the problem with gtpin's use of dlmopen. For Aaron and Keren to work in with code from develop, they need this fix to avoid LD_AUDIT. This moves us in the direction where our implementation is modern: it is organized for LD_AUDIT but has a workaround for platforms where LD_AUDIT has problems. Do you have any objections? 

This commit can be exercised more fully after being accepted, just like we did with the LD_AUDIT commit (exposing it to gtpin's dlmopen). That wouldn't have happened if LD_AUDIT were in Jonathon's private branch because he didn't have access to the system at Argonne in JLSE where it failed.

mwkrentel:
As long as you're putting it into develop, I think you should go ahead
and merge it.

PR #331, opened 2020-09-26
No unwind implementation

martyitz:
I think I've done the right git-stuff, and would like to see the pull into master before the release.

     Thanks,
          Marty

### DIFF
--- a/doc/man/hpcrun.tex
+++ b/doc/man/hpcrun.tex
@@ -218,6 +218,12 @@ If this option is given, \Prog{hpcrun} will record all elements of a recursive c
 Note: When you use the \Prog{RETCNT} sample source this option is enabled automatically
 to gather accurate counts.
 
+ \item[\Opt{-nu}, \Opt{--no-unwind}]
+Suppress unwinding of callstacks.
+Normally as \Prog{hpcrun} collects data against full callstacks at the time of each event.
+If this option is given, \Prog{hpcrun} will skip unwinding, and generate data with only
+the leafPC's.
+
 \item[\Opt{-t}, \Opt{--trace}]
 Generate a call path trace in addition to a call path profile.
 
--- a/src/tool/hpcrun/fnbounds/fnbounds_client.c
+++ b/src/tool/hpcrun/fnbounds/fnbounds_client.c
@@ -487,9 +487,9 @@ launch_server(void)
 
   if (pid < 0) {
     //
-    // fork failed
+    // clone failed
     //
-    EMSG("FNBOUNDS_CLIENT ERROR: syserv launch failed: fork failed");
+    EMSG("FNBOUNDS_CLIENT ERROR: syserv launch failed: clone failed");
     return -1;
   }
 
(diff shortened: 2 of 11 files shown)


mwkrentel (approved):

mwkrentel:
LGTM

@jmellorcrummey  With your permission, I'd like to change one thing.

For the boolean environ variables like HPCRUN_NO_UNWIND, I'd like to
change this to: HPCRUN_NO_UNWIND must be set to a non-zero value to
turn off unwinding.

```
  char *foo = getenv("HPCRUN_NO_UNWIND");
  if (foo != NULL){
    hpcrun_no_unwind = true;
  }
```

That is, HPCRUN_NO_UNWIND=0 should NOT turn off unwinding.
I think this has less shock value to users.

I know we violate this in other places, IIRC HPCRUN_TRACE, and I've
never liked that.

So, Marty, if you could change this one, I'll try to track down and
fix the other variables.

Btw, I tried this end-to-end in the viewer, and it's interesting that
you still get loops and inline sequences, at least within the
innermost function.

-----------

Btw, somehow I clicked on the wrong 'review' link.  I was trying to
add my review, but somehow I deleted the request for review for John
and I think it was Keren.  My apologies, a clear case of incompetence
over malice. :-)

jmellorcrummey:
I agree with Mark's suggestion. Setting HPCRUN_NO_UNWIND to a non-zero value is a more intuitive way to control things than simply seeing that the environment variable is set.

I also approve.

mwkrentel:
I also meant to ask, do we want to call this 'no-unwind' or 'flat'.
I suppose I could go either way, but I think we should decide one way
or the other now.

martyitz:
So, what exactly do you want me to do?

Do I change -nu and --no-unwind to -fl and --flat-profile?

The hpcrun_no_unwind is a bool, not an int.  What should I do with it?

These changes all seem pointless,, but your wish is my command.

	M

mwkrentel:
I was looking for @jmellorcrummey to step in and choose.

jmellorcrummey:
replace -nu and —no-unwind with  —flat and -f.

the environment variable should be HPCRUN_FLAT=0 or 1 for a bool. we don’t parse true and false for anything.

mwkrentel:
Well, -f is already in use for process fraction.
So, let's just go with  -flat  and  --flat.

mwkrentel:
@jmellorcrummey   I'm going to go ahead and merge this as is.
We can revisit two things after merging.

1. I'm certainly going to adjust the treatment of the environ variable
so that HPCRUN_FOO=0 does not turn on FOO.  That's for no-unwind,
trace and anything else I can find.

2. We may want to revisit no-unwind vs flat before the next release.
-nu isn't very mnemonic and we've previously called this flat.
As long as we settle this before the release, that's ok.
But I don't want to hold this up until then.

jmellorcrummey:
> I'm certainly going to adjust the treatment of the environ variable
> so that HPCRUN_FOO=0 does not turn on FOO. That's for no-unwind,
> trace and anything else I can find.

Agreed.

> We may want to revisit no-unwind vs flat before the next release.
> -nu isn't very mnemonic and we've previously called this flat.
> As long as we settle this before the release, that's ok.
> But I don't want to hold this up until then.

-flat is better. as you said, this is what we had previously for MIPS.

PR #332, opened 2020-09-29
Enhance hpcrun_reify_metric_set to eagerly allocate metrics_kind

Jokeren:
The existing hpctoolkit **doesn't** have the bug I mentioned in the meeting last week. Originally, I thought hpcrun incorrectly attributes metrics if there are multiple metrics kind because the allocation interface `hpcrun_reify_metric_set` does not consider the case when different `metric_id`s are passed for the same `cct_id`. 

However, the bug actually was resolved by myself back to 2019 in this commit:

https://github.com/HPCToolkit/hpctoolkit/commit/0a4707c8be275fd5427269c2e89c4c1501921932

This commit guarantees the correctness of metrics attribution by lazy initialization of all the metric kinds. The following code change propose a solution to early initialize metric kinds directly in `hpcrun_reify_metric_set`. Hope this change could make the code clearer. 

If needed, I can remove the previous code and test if early initialization is enough.


jmellorcrummey:
I read the code and approve of the change. This indeed makes the allocation of metrics clearer.

PR #335, opened 2020-09-29
Demand new thread data for gpu_trace_stream

dejangrubisic:
Ensure that cpu_thread never becomes a prefix of gpu thread.

### DIFF
--- a/src/tool/hpcrun/gpu/gpu-trace.c
+++ b/src/tool/hpcrun/gpu/gpu-trace.c
@@ -345,12 +345,16 @@ gpu_trace_stream_acquire
  void
 )
 {
+  bool demand_new_thread = true;
+  bool has_trace = true;
+
   thread_data_t *td = NULL;
 
   int id = gpu_trace_stream_id();
 
   // XXX(Keren): This API calls allocate_and_init_thread_data to bind td with the current thread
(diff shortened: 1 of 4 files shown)


jmellorcrummey (approved):

jmellorcrummey:
@mxz297 @laksono this commit should address the out of order trace you reported on AMD with floyd warshall. the problem there was that a cpu thread finished and a gpu thread was appended. the gpu clock was skewed a bit.

laksono:
+1
@dejangrubisic @jmellorcrummey Personally, Instead of having `bool` parameters, I prefer to have `int` flag so that we can add more options in the future. With boolean, we stuck with either true or false.
Hence, I propose something like:
```
hpcrun_threadMgr_data_get_safe(int id, cct_ctxt_t* thr_ctxt, thread_data_t **data, int option);
hpcrun_threadMgr_data_get(int id, cct_ctxt_t* thr_ctxt, thread_data_t **data, int option);
```
Instead of:
```
hpcrun_threadMgr_data_get_safe(int id, cct_ctxt_t* thr_ctxt, thread_data_t **data, bool has_trace, bool demand_new_thread);
hpcrun_threadMgr_data_get(int id, cct_ctxt_t* thr_ctxt, thread_data_t **data, bool has_trace, bool demand_new_thread);
```

PR #336, opened 2020-09-30
replace gotcha with objcopy redefine-sym to intercept dl_iterate_phdr in hpcrun

mwkrentel:
Replace using gotcha in hpcrun to intercept dl_iterate_phdr() in
libunwind with objcopy to redefine the function name to
hpcrun_real_dl_iterate_phdr().  This is more targeted and robust
and eliminates needing to use gotcha in hpcrun.

For now, we'll leave the gotcha option in configure and simply comment
out the test.  Looking ahead, we may end up using gotcha for something
else.

This should fix the accidental breakage for Kull by statically linking
libunwind into hpcrun.

Remove the dual support for libunwind and lzma being compiled with or
without -fPIC and now always require -fPIC.

Going forward with the hiding libs work, it's going to be too much
trouble to maintain both alternatives.  The non-fpic case would
continue to require gotcha, whereas in the fpic case, we can remove
gotcha altogether.

----------

Some notes:

1. Hpcrun only computes a replacement for dl_iterate_phdr() in the
dynamic case.  But libunwind doesn't distinguish between the static
and dynamic cases and needs a real answer (not just one module) even
in the static case or else produces 100% bad unwinds.

So, fallback to calling the real dl_iterate_phdr in the static case.
But a static program probably doesn't call dlopen/dlclose, so there is
not much chance of deadlock.

2. Please don't merge unless it's a fast-forward merge.  I can easily
rebase if master receives more commits and I'd prefer to keep linear
history.

3. I'd also prefer to keep this as two separate commits.  The first
commit doesn't change anything except require static + fpic.
So, don't squash.

ping @jmellorcrummey   @blue42u


src/tool/hpcrun/extern-real/dl-iterate.c
```diff
@@ -0,0 +1,83 @@
...
+
+int
+hpcrun_real_dl_iterate_phdr (
+  int (*callback) (struct dl_phdr_info * info, size_t size, void * data),
+  void *data
+)
+{
+#ifdef HPCRUN_STATIC_LINK
+  // statically, we have no replacement, so call the real version
+  return dl_iterate_phdr(callback, data);
```
blue42u, line 76: > But libunwind doesn't distinguish between the static
and dynamic cases and needs a real answer (not just one module) even
in the static case or else produces 100% bad unwinds.

I'm going to need a bit more explanation here. Correct me if I'm off base, but wouldn't this mean that the loadmap generated at `hpclink` time is incorrect? My understanding is the point of static binaries is to not load any other libraries (unless explicitly requested).

We have the requirement from other code (somehow), but I'd prefer not to add more reasons to force `-ldl` on the `hpclink` line if possible.

If the offsets just need relocated, then we might be able to handle that with a little math from `getauxval(AT_PHDR)`. Otherwise this sounds like another latent bug that we just never happened to notice.

Either way, I don't expect this to be solved in this PR, but it should be investigated.

blue42u (approved):
> Remove the dual support for libunwind and lzma being compiled with or
without -fPIC and now always require -fPIC.

Good choice. Mind sending a reminder on the mailing list once this is merged?

I'd also like to request a subsequent merge to `develop` to keep things as consistent as possible. I'll communicate with Yumeng and handle the rebase for `prof2`.

> 2. Please don't merge unless it's a fast-forward merge. I can easily
rebase if master receives more commits and I'd prefer to keep linear
history.

Or use "Rebase and merge" in the GUI. It does mess with the commiter data, but the history stays linear.
Now if only https://github.com/isaacs/github/issues/1017 was implemented and we could have the even better semi-linear history...

---
Otherwise without testing looks reasonable, remaining comments alongside relevant code below.

src/extern/libunwind/Makefile.am
```diff
@@ -59,7 +59,7 @@ NEW_LIB = libunwind.a
 
 OBJCOPY = objcopy
 
-SYMBOL_NAMES = mmap munmap
+SYMBOL_NAMES = mmap munmap dl_iterate_phdr
 
 REDEFINE_ARGS = $(foreach name,$(SYMBOL_NAMES),--redefine-sym $(name)=hpcrun_real_$(name))
```
jmellorcrummey, line 64: I think that this yields a bad name: hpcrun_real_dl_iterate_phdr, which isn't the real function; it is a replacement.

For mmap and munmap, the hpcrun_real code implements the real function.

src/tool/hpcrun/extern-real/dl-iterate.c
jmellorcrummey: @blue42u the load map for a static binary is simple: the executable only, which contains everything.

src/tool/hpcrun/extern-real/dl-iterate.c
```diff
@@ -0,0 +1,83 @@
...
+//----------------------------------------------------------------------
+
+#define _GNU_SOURCE  1
+
+#include <sys/types.h>
+#include <link.h>
+#include "loadmap.h"
+
+int
+hpcrun_real_dl_iterate_phdr (
```
jmellorcrummey, line 69: I think that we could have the ld_audit layer provide a better implementation than what we do in hpcrun now. Wouldn't that work for the static case too?

jmellorcrummey (changes requested):
While this is not an earthshaking request, I think that the function named hpcrun_real_dl_iterate_pheader is mistakenly named. changing the macros a bit so that you could just replace dl_iterate_pheader with hpcrun_dl_iterate_pheader. For the dynamic case, this provides the functionality hpcrun needs in an async signal safe way. for the static case, simply replacing it with dl_iterate_pheader seems risky.

a larger change to contemplate: replacing hpcrun_dl_iterate_pheader with a function that implements the necessary user callbacks using the copy of the load map maintained in audit space. would that require turning off signals in audit space while we maintain the table so that we can never get interrupted and see the tables in an inconsistent state? better would be to use a non-blocking data structure in audit space :-) 

what does @blue42u think?

mwkrentel:
> Or use "Rebase and merge" in the GUI. 

I know what to do from the command line, but John and I have both had
trouble with the github GUI.  Anyway, you understand what I want, right?
Linear history and separate commits (no squash).

mwkrentel:
> the load map for a static binary is simple: the executable only,
> which contains everything.

So you would think.  I tried dl_iterate_phdr() in a static program on
arm1 and got this (using the example from the man page).

```
name=  (6 segments)
         header  0: address=  0x400000
         header  1: address=  0x588b18
         header  2: address=  0x400190
         header  3: address=  0x588b18
         header  4: address=     (nil)
         header  5: address=  0x588b18

name=linux-vdso.so.1  (4 segments)
         header  0: address=0xffff8dc32000
         header  1: address=0xffff8dc32738
         header  2: address=0xffff8dc322b8
         header  3: address=0xffff8dc32674
```

Running a static program on arm1, if I use the real dl_iterate_phdr()
in hpcrun, then I get the right answer.  If I use the hpcrun one-item
substitute, then I get 100% bad unwinds.

Anyway, we've never correctly supplied a substitute for the static
case.  We could review this, but that would be a new feature, so I'd
prefer to leave that to future work.

blue42u:
> But libunwind doesn't distinguish between the static
and dynamic cases and needs a real answer (not just one module) even
in the static case or else produces 100% bad unwinds.

> If I use the hpcrun one-item
substitute, then I get 100% bad unwinds.

The 100% (not 50% or 99%) is what makes me suspicious that the values returned by (now) `hpcrun_real_dl_iterate_phdr` are faulty in the static case. In which case the data stored by `fnbounds_init` is just plain wrong, or is missing something that wasn't added when `hpcrun_iterate_loadmap` was first written.

I don't think that should delay this PR, but its something that should probably be investigated if any more static issues come up. Especially if we have deadlocks with (effectively) a static dlstress test.

src/tool/hpcrun/extern-real/dl-iterate.c
blue42u: The auditor just shoves the callbacks to libhpcrun, so an implementation in the auditor would be about the same. Not that there couldn't be improvements (atomics), but being in audit-space won't provide a significant benefit.

Currently the auditor isn't enabled in the static case, the real version isn't possible for obvious reasons. The fake would be viable, but I'd rather keep it dynamic-only at least until the specifics of the auditor's role are settled.

IMO, the static case is gradually becoming less of a required use case. Shared libraries are easier on both the sysadmins and developers, and the performance benefits could be retained by preloading data based on `ldd` or using Spindle (or even better, a distributed read-only filesystem cache based on FUSE). As the case where we're both blindfolded and have one arm tied, the growing maintenance cost (from other needed refactors) needs to evaluated against the actual benefit.

(That'll be my opinion until the next big ARM supercomputer decides its going to be static-only. Which I sincerely hope doesn't happen, and won't as long as GPUs are still useful and ARM is still unstable. And maybe containers.)

PR #337, opened 2020-10-02
Delay auditor notifications until libhpcrun is certainly ready

blue42u:
If libhpcrun initializes early, the auditor may witness its initialization and assume its prepared to receive notifications before it actually is (via a recursive initialization attempt).
This patch instead assumes that the mainlib will notify the auditor when it is prepared, and thus delays notifications until that point.

The middling `connecting` state is still needed to prevent (some) recursive initializations.

Fixes issues with PeleC and other applications that spawn threads in their init constructors.

@mxz297


jmellorcrummey:
I have finally read and understood why the additional calls to mainlib_connected are a problem in the auditor code. I approve the pull request.

PR #345, opened 2020-11-17
PLT call performance fix

mxz297:
Tested on gpu (standard x86 PLT), ufront (security extension x86 PLT), and llnl (power).

### DIFF
--- a/src/tool/hpcrun/audit/auditor.c
+++ b/src/tool/hpcrun/audit/auditor.c
@@ -68,6 +68,8 @@
 #define GOT_resolve_offset 2
 #elif defined(HOST_CPU_PPC)
 #define GOT_resolve_offset 0
+#elif defined(HOST_CPU_ARM64)
+#define GOT_resolve_offset 2
 #else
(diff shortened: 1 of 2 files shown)


blue42u (approved):
Few trailing whitespace errors, otherwise code looks good to me.

For something as expert-mode as this I don't think we need the `-dapco` shorthand, `--disable-audit-plt-call-opt` is explanatory enough.

jmellorcrummey (commented):
Should we go ahead and add GOT offset for ARM? It just occurred to me that we won't be executing the buggy code in the dl_runtime_profile on ARM once we replace the PLT resolver.

mxz297:
@jmellorcrummey That's worth testing. I am not sure this change can completely remove the execution `dl_runtime_profile`. I will test it.

mxz297:
@jmellorcrummey Your intuition is correct. Now the auditing layer runs fine on arm.

jmellorcrummey (approved):

PR #350, opened 2020-11-25
adjust vdso handling in auditor.c

jmellorcrummey:
fix issues with vdso entry presented to hpctoolkit
- fix start address
- fix end address
- fix dpi_addr

### DIFF
--- a/src/tool/hpcrun/audit/auditor.c
+++ b/src/tool/hpcrun/audit/auditor.c
@@ -147,8 +147,6 @@ static auditor_exports_t exports = {
 
 static struct buffered_entry_t* obj_update_list = NULL;
 
-static const bool vdso_load_module = true;
-
 
 
 //******************************************************************************
--- a/src/tool/hpcrun/audit/auditor.c
+++ b/src/tool/hpcrun/audit/auditor.c
@@ -94,9 +94,6 @@
 #error "GOT resolver index for the host architecture is unknown"
 #endif
 
-#define IS_VDSO  true
-#define NOT_VDSO false
-
 
 
 //******************************************************************************
@@ -116,6 +113,12 @@ struct buffered_entry_t {
 
   struct buffered_entry_t* next;
 } *buffer = NULL;
+
+
+enum audit_open_flags {
+  AO_NONE = 0x0,
+  AO_VDSO = 0x1,
+};
 
 
 
@@ -144,6 +147,8 @@ static auditor_exports_t exports = {
 
 static struct buffered_entry_t* obj_update_list = NULL;
 
+static const bool vdso_load_module = true;
+
 
 
 //******************************************************************************
@@ -162,7 +167,7 @@ static ElfW(Addr) * get_plt_got_start(ElfW(Dyn) * dyn_init) {
 
 
 // Helper to call the open hook, when you only have the link_map.
-static void hook_open(uintptr_t* cookie, struct link_map* map, bool is_vdso) {
+static void hook_open(uintptr_t* cookie, struct link_map* map, enum audit_open_flags ao_flags) {
   // Allocate some space for our extra bits, and fill it.
   auditor_map_entry_t* entry = malloc(sizeof *entry);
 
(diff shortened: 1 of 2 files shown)


jmellorcrummey:
This needs another commit to fix the fake auditor to address the same issues.

src/tool/hpcrun/audit/auditor.c
```diff
@@ -237,11 +240,22 @@ static void hook_open(uintptr_t* cookie, struct link_map* map) {
...
+    entry->end = (void*)map->l_addr + (end - start);
+  } else {
+    entry->start = (void*)map->l_addr + start;
+    entry->end = (void*)map->l_addr + end;
+  }
 
   // Since we don't use dl_iterate_phdr, we have to reconsitute its data.
-  entry->dl_info.dlpi_addr = (ElfW(Addr))map->l_addr;
+  if (is_vdso) {
+    entry->dl_info.dlpi_addr = (ElfW(Addr))(map->l_addr - start);
```
blue42u, line 254: Please modify `map->l_addr` here to the proper value, I believe it and `dl_info.dlpi_addr` are supposed to be in sync. Also that way I don't have to remember which is the "right" one when calculating offsets.

src/tool/hpcrun/audit/auditor.c
```diff
@@ -94,6 +94,9 @@
 #error "GOT resolver index for the host architecture is unknown"
 #endif
 
+#define IS_VDSO  true
+#define NOT_VDSO false
```
blue42u, line 98: I'd prefer to keep these written as their more obvious `true` and `false`, unless there might be other options (in which case I would prefer an `enum` of flag bits). In my experience macros tend to complicate debugging and code readability, and so should only be used sparingly when no other option is available.

src/tool/hpcrun/audit/fake-auditor.c
```diff
@@ -275,15 +269,27 @@ int update_shadow_dl(struct dl_phdr_info* map, size_t sz, void* args_vp) {
...
   }
+
   entry->entry.start = (void*)map->dlpi_addr + start;
   entry->entry.end = (void*)map->dlpi_addr + end;
 
+  entry->entry.path = NULL;
+  if(args->index == 0 && map->dlpi_name[0] == '\0')
+    entry->entry.path = realpath((const char*)getauxval(AT_EXECFN), NULL);
+  else if (entry->entry.start == getauxval(AT_SYSINFO_EHDR))
+    entry->entry.path = realpath(vdso_path, NULL);
```
blue42u, line 280: The vDSO condition should still be stronger (more certain) than the main executable condition, so these two should be swapped.

blue42u (changes requested):

src/tool/hpcrun/audit/auditor.c
jmellorcrummey: I hate seeing true/false arguments because they don't make any sense without looking at the a function's argument definition. How about this:

`static const bool vdso_load_module = true;`

pass as argument: vdso_load_module, ! vdso_load_module

src/tool/hpcrun/audit/fake-auditor.c
jmellorcrummey: done

src/tool/hpcrun/audit/auditor.c
blue42u: I would argue the `static const` is worse, I can't tell at a glance that `vdso_load_module` is immutable and it acts like a real variable (`&vdso_load_module` makes no sense but is not malformed). You usually have to check the function's declaration for any real work anyway, and gdb will give you the argument names as part of the backtrace.

The other option I agree with is an enum of flag bits, it has the syntax additions of a macro and the debuggability of a standard type (gdb will break it out for you: `#0  foo (x=(X_1 | X_2)) at x.c:10`).
```c
enum audit_open_flags {
  AO_NONE = 0x0,
  AO_VDSO = 0x1,
};
...
static void hook_open(uintptr_t* cookie, struct link_map* map, enum audit_open_flags flags) {
  // is_vdso = flags & AO_VDSO
...
```
Usage: `AO_VDSO`, `AO_NONE`.

jmellorcrummey: Of course, all of this is predicated that you can examine the code under the debugger, which you can't. I want to write it for readability, which is the only way that one can debug auditor code. 

I will switch to these flags, which read a lot like what I originally had.

src/tool/hpcrun/audit/auditor.c
jmellorcrummey: resolved.

jmellorcrummey:
I tested with quicksilver on both x86_64 and power and I tested with a dl_iterate_phdr test that allows me to directly compare the numerical values for vdso dlpi_addr, start, and end. Everything is working smoothly.

src/tool/hpcrun/audit/auditor.c
```diff
@@ -237,8 +243,13 @@ static void hook_open(uintptr_t* cookie, struct link_map* map) {
...
   }
-  entry->start = (void*)map->l_addr + start;
-  entry->end = (void*)map->l_addr + end;
+
+  // load module.  we must adjust it by start so that it has the proper
+  // base address value for the subsequent calculations.
+  if (ao_flags & AO_VDSO) map->l_addr -= (long) start;
+
+  entry->start = (void*)map->l_addr + (long) start;
+  entry->end = (void*)map->l_addr + (long) end;
```
blue42u, line 252: Last comment: is there a reason these are `long` and not `intptr_t` (or better `uintptr_t`)? `long` is not necessarily large enough to hold a pointer-sized value.

Otherwise everything looks fine.

blue42u (approved):

src/tool/hpcrun/audit/auditor.c
jmellorcrummey: change to intptr_t. the arithmetic didn't work out with uintptr_t.

PR #357, opened 2020-12-13
Fix issue 356 -- CPU time is overcounted when threads sleep.

martyitz:
The code used the delta in REALTIME as the value of CPUTIME when profiling
by the latter.  If a thread was sampled at T-epsilon, and then went to sleep
until T2 + epsilon, the time between T2 and T would be incorrectly said
the be CPU time.

The fix is to read the CPU time for the delta when profiling CPUTIME,
and the real time for the delta when profiling REALTIME.

While I was in there, I deleted the various tortuous bits of code that were
needed to support BlueGene, which is now completely obsolete.  I also deleted
the profiling of WALLCLOCK, which was just an alias for CPUTIME, and profiling
by the undocumented timer ITIMER.

Reworked the hpcrun -l message to correctly describe the timer options.


jmellorcrummey (changes requested):
Now that there is no itimer support, any message that says "itimer" should say "Linux timer".

Now that there is no itimer support, when ENABLE_REALTIME_CLOCK is not set, the sample source should not be included in the list of sample sources. There should be an #ifdef ENABLE_REALTIME... around the inclusion of the itimer sample source in the sample source list in the ss-list* file.

With the above change, the error messages when  ENABLE_REALTIME* is not defined can be deleted.

martyitz:
Making the change requested to ss-list.h turns off the timers.  ENABLE_REALTIME_CLOCK is defined compiling
timer.c, but not when compiling the code that supports hpcrun -L or hpcrun -e CPUTIME.

mwkrentel:
@jmellorcrummey 
First, I'm ok with removing ITIMER from the sample sources that we
support.  We actually needed it for Blue Gene, but now that's gone,
and itimer has a lot of problems, including it's not thread-specific
(except it was on blue gene).

@martyitz 
I guess this will do what you intend, but it obscures the two main
points of your patch: (1) remove itimer, you're treating this as
"while I'm here, I'll do this cleanup", but actually it's much larger
than ... (2) fixing CPUTIME.

I think it would be cleaner to do, in either order, but separate
commits:

(1) fix CPUTIME and leave ITIMER alone.
(2) remove ITIMER.

But for (2), here's how you should do it:

Now, itimer.c assumes that ITIMER always exists and REALTIME/CPUTIME
may or may not exist (which was true because of blue gene).

Removing ITIMER and Blue Gene leaves you with: there is no itimer, so
the entire file is only for the case that REALTIME exists.  I'd rip
out the #ifdef ENABE_CLOCK_REALTIME, still test for this is configure,
but then either compile the entire file or not.  That is, replace the
#ifdef's with a single yes/no at the Makefile level.

Btw, REALTIME will always exist.  But I'd suggest keep the configure
test and pass this to Makefile.

martyitz:
I have committed my latest.  I don't know how to update the pull request.

martyitz:
I removed BG and ITIMER according to the instructions from John.
I have pushed my latest to cputime.  Someone should pull it into develop.

mwkrentel:
Ok, let's put Marty out of his misery and let him move on to something
else productive. :-)

To be clear, my objection is combining two patches where the "main"
patch is tiny and the "while I'm here" patch is much larger.

I'll make a pass over itimer.c to clean up the #ifdefs.

> I have committed my latest. I don't know how to update the pull request.

You can push commits to the same branch and they'll be added
automatically.

martyitz:
Thank you.

PR #362, opened 2020-12-24
Rework configure for spack rocm arguments.

mwkrentel:


mwkrentel:
These are the changes that I need to support building with spack.
It allows configuring with rocm either all-in-one or by separate
spack packages, or a mixture.

      --with-rocm=path    (for all-in-one /opt/rocm)
      --with-rocm-hip     (for individual packages)
      --with-rocm-dbgapi
      --with-rocm-tracer

Be sure to test that it runs with a simple rocm program.  I've
replicated the configure/makefile actions as before, but I haven't
done a run test.

I suggest squashing the merge because some of the intermediate commits
will build but not run.

mwkrentel:
ping @mxz297 for review.  no rush.

mxz297 (approved):
Looks good to me.

PR #363, opened 2021-01-06
Level0 wrappers

mxz297:
Isolate level 0 public API wrappers to a separate libhpcrun_level0.so. Also fixed regression on rocm and nvidia platforms.


jmellorcrummey (approved):
LGTM

PR #366, opened 2021-01-08
Opencl wrappers

mxz297:
Move OpenCL wrappers to libhpcrun_opencl.so. This also includes fixes to Level0 support for a bug that I accidentally encountered. I also compile the code with both `--with-level0` and `--with-opencl` at configure time. The same build works for dpcpp code using either level0 backend or opencl backend, meaning `SYCL_BE=PI_LEVEL_ZERO hpcrun -e gpu=level0 ...` and `SYCL_BE=PI_OPENCL hpcrun -e gpu=opencl ...` both works.

@jmellorcrummey I don't have an environment to test GTPin support ready. Can you test this branch with GTPin?


jmellorcrummey:
I will check this out and test it at Argonne using GT-Pin.

jmellorcrummey (approved):
Tested at ANL with GT-Pin. Works as expected.

PR #369, opened 2021-01-14
Fix typo in hpcrun script for setting roctracer path

mxz297:
Current develop branch will fail to find a library from roctracer due to a typo in the hpcrun script.


jmellorcrummey (approved):
LGTM.

PR #370, opened 2021-01-18
Fix issue 321 - problem with dynamic library unmap notification 

### LINKED ISSUE #321: hpcrun loses track of dlopen'd object
A test program that opens one shared object, then closes it and opens a second, and then reloads
the first, loses track of the second library.  The test is constructed so that the second library
gets loaded at the same address as the first.  The symptom of the bug is that the addresses
from the second library are treated as coming from an unknown function in an unknown module.
The reload of the first library, now at a different address, is handled correctly.

The bug is absent in the  2019.08 and 2020.03 bits;  it appears first in 2020.06, and persists in 2020.08 and in master.

I have a relatively simple reproducer.

When run with hpcrun -dd UW_RECIPE_MAP, the following errors are found:
      3 [47038, N]: UW_RECIPE_MAP: build_intervals: fcn range 0x7f7e9188a710 to 0x7f7e9188a7f0: error -10
      4 [47038, N]: UW_RECIPE_MAP: build_intervals: fcn range 0x7ffea11b9600 to 0x7ffea11b9b50: error -10
      5 [47038, N]: UW_RECIPE_MAP: build_intervals: fcn range 0x7f7e913bf7e0 to 0x7f7e913bf820: error -10

jmellorcrummey:
fix #321. in short, unmap notification to the unwind recipe map occurred too late - after a shared library's representation as a dso_info object in the loadmap was cleared. remedy this by performing unmap notifications before discarding the dso_info for a shared library in the loadmap.

### DIFF
--- a/src/tool/hpcrun/loadmap.c
+++ b/src/tool/hpcrun/loadmap.c
@@ -417,6 +417,50 @@ hpcrun_loadmap_moveToBack(load_module_t* lm)
 #endif
 
 
+static void
+hpcrun_loadmap_dump_dl_phdr_info(struct dl_phdr_info *x)
+{
+  fprintf(stderr, "\t phdr_info %p\n", x);
+  if (x) {
+    fprintf(stderr, "\t\t dlpi_addr 0x%lx\n", x->dlpi_addr);
+    fprintf(stderr, "\t\t dlpi_name %s\n", x->dlpi_name);
+    fprintf(stderr, "\t\t dlpi_phnum %d\n", x->dlpi_phnum);
+    fprintf(stderr, "\t\t dlpi_adds %llu\n", x->dlpi_adds);
+    fprintf(stderr, "\t\t dlpi_subs %llu\n", x->dlpi_adds);
+    fprintf(stderr, "\t\t dlpi_tls_modid %ld\n", x->dlpi_tls_modid);
+    fprintf(stderr, "\t\t dlpi_tls_data %p\n", x->dlpi_tls_data);
+  }
+}
+
+
+static void
+hpcrun_loadmap_dump_dso_info_t(dso_info_t* x)
+{
+  fprintf(stderr, "\t dso_info_t %p\n", x);
+  if (x) {
+    fprintf(stderr, "\t\t name %s\n", x->name);
+    fprintf(stderr, "\t\t start_addr %p\n", x->start_addr);
+    fprintf(stderr, "\t\t end_addr %p\n", x->end_addr);
+    fprintf(stderr, "\t\t map_size %lu\n", x->map_size);
+    fprintf(stderr, "\t\t nsymbols %lu\n", x->nsymbols);
+    fprintf(stderr, "\t\t is_relocatable %p\n", x->is_relocatable);
+  }
+}
+
+
+static void
+hpcrun_loadmap_dump_load_module_t(load_module_t* x)
+{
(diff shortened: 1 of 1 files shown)


jmellorcrummey:
@martyitz FYI: this pull request fixes your issue #321. You can try the code in branch fix_issue_321 if you want.

blue42u (changes requested):
So, basically this always triggered this line at the top of `uw_recipe_map_notify_unmap`, causing the entire call to effectively no-op and not clear out the unwind recipes properly:
https://github.com/HPCToolkit/hpctoolkit/blob/7c042fc8b3382d88c2642d37c845b10edc3c8254/src/tool/hpcrun/unwind/common/uw_recipe_map.c#L576

I'm not going to ask how long it took to hunt this one down. Assuming it works, great!

---

For the new debugging functions, IMO they should either:
 - a) be condensed into `hpcrun_loadmap_print` instead of adding `hpcrun_loadmap_dump`,
 - b) be printed (via a `TMSG(LOADMAP)` guard) as part of the normal operations,
 - or c) be removed altogether since gdb can give the same data (although much more manually).

Mostly for the purpose of reducing the amount of mess near delicate core pieces of functionality.

jmellorcrummey:
> So, basically this always triggered this line at the top of `uw_recipe_map_notify_unmap`, causing the entire call to effectively no-op and not clear out the unwind recipes properly:
> 
> https://github.com/HPCToolkit/hpctoolkit/blob/7c042fc8b3382d88c2642d37c845b10edc3c8254/src/tool/hpcrun/unwind/common/uw_recipe_map.c#L576

You are correct. Previously, the guard at the top of  uw_recipe_map_notify_unmap saw that dso_info was NULL and skipped the operation. 

> I'm not going to ask how long it took to hunt this one down.

It only took an hour or so to identify and fix after writing hpcrun_loadmap_dump. Without the new dumping code, it would have required a lot more tedious examination of internal state to pinpoint the root cause of the failure.

> For the new debugging functions, IMO they should either:
> 
> * a) be condensed into `hpcrun_loadmap_print` instead of adding `hpcrun_loadmap_dump`,

These two debugging functions serve different purposes. hpcrun_loadmap_print prints only the association between load module name and load module id, which can be useful when using the debugger to examine data in CCT nodes.

hpcrun_loadmap_dump serves a different purpose: it provides a complete dump of the loadmap, which is useful when debugging problems with the visibility of dynamic libraries.

> * b) be printed (via a `TMSG(LOADMAP)` guard) as part of the normal operations,

IMHO, hpcrun_loadmap_dump prints much more information than one wants to see when tracing operations. Also, the place where I want to use such operations varies with the thing being debugged. For resolving this problem, I dumped the loadmap not when it was modified. Rather, I dumped it when hpcrun failed to compute a normalized IP to understand why. The helper functions are there to accelerate tedious examination in a debugger.  If it were dumped each time each of M load modules were loaded or unloaded, it would generate O(M^2) output.

> * or c) be removed altogether since gdb can give the same data (although much more manually).

While it is true that one can observe the same data with gdb, dumping each entry requires dumping through three independent structure pointers. Doing that manually for a loadmap with even tens of load modules is beyond tedious. 

> 
> Mostly for the purpose of reducing the amount of mess near delicate core pieces of functionality.

I sympathize with your goal. However, as the person who debugs hpcrun, IMO poking around in the dark without dumping contents at appropriate points is a frustrating waste of time.

mxz297 (approved):

PR #372, opened 2021-01-21
introduced how arg in device_finalizer_fn

dejangrubisic:
Adaptation to Keren's change of finalizer function, and deleting gpu_operation_channel_set_await as unnecessery

### DIFF
--- a/src/tool/hpcrun/gpu/gpu-trace.c
+++ b/src/tool/hpcrun/gpu/gpu-trace.c
@@ -361,10 +361,7 @@ gpu_trace_record
     gpu_trace_channel_set_process(channel_set);
 
   }
-
   gpu_trace_channel_set_process(channel_set);
-  gpu_trace_channel_set_await(channel_set);
-
   gpu_trace_channel_set_release(channel_set);
   return NULL;
 }
--- a/src/tool/hpcrun/gpu/gpu-trace.c
+++ b/src/tool/hpcrun/gpu/gpu-trace.c
@@ -334,7 +334,8 @@ gpu_trace_init
 void
 gpu_trace_fini
 (
- void *arg
+ void *arg,
+ int how
 )
 {
   PRINT("gpu_trace_fini called\n");
--- a/src/tool/hpcrun/gpu/gpu-trace.c
+++ b/src/tool/hpcrun/gpu/gpu-trace.c
@@ -65,6 +65,7 @@
 #include <lib/prof-lean/stdatomic.h>
 
 #include <hpcrun/cct/cct.h>
+#include <hpcrun/utilities/hpcrun-nanotime.h>
 #include <hpcrun/control-knob.h>
 #include <hpcrun/thread_data.h>
 #include <hpcrun/threadmgr.h>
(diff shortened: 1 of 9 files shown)


dejangrubisic:
This should improve tracing speed. I have checked out Lagos and the realtime was 11.969s while the current develop is 17.318s.
Produced database looks as expected in traceviewer.

Jokeren:
I remembered it is ~11.9s vs ~14.0s something.

Please attach the results with the following three configurations.

- develop with -t
- this commit with -t
- this commit without -t

dejangrubisic:
This time I got:
    develop with -t -> 10.457s
    this commit with -t -> 8.727s
    this commit without -t -> 7.524s

jmellorcrummey (commented):
I don't understand what this commit means. There aren't any calls to gpu_operation_channel_set_await anymore. Are threads just spinning hard all the time? Or, do they sleep at some point?

dejangrubisic:
This commit:
1. Contains a change in definitions of gpu_trace_fini (adding how arg),
2. In current develop gpu_operation_channel_set_await was already removed from the loops. This commit removes it after the loop, which resulted in better performance.
3. Moves control knob_init after hpcrun_registered_sources_init, as otherwise, it wouldn't pick user-defined knobs

src/tool/hpcrun/sample_sources_registered.c
```diff
@@ -145,6 +145,8 @@ hpcrun_registered_sources_init(void)
     METHOD_CALL(ss, init);
     TMSG(SS_COMMON, "sample source \"%s\": init", ss->name);
   }
+  
+  control_knob_init();
```
Jokeren, line 149: I am ok with all the changes.

Jokeren (approved):

jmellorcrummey (changes requested):
I don't think that control_knob_init should be called from sample_sources_init. Why do you think it should?

Why did you add an argument to gpu_trace_fini if it is always called with a 0 for how. It should never be passed a number. The number should be replaced with the named constant from monitor.h.

Why can a call to channel_set_await just be deleted?

dejangrubisic:
I instead of hard-spinning on each channel I inserted nanosleep() of for total sleep of 5us. Compared with hard-spinning user time is much lower.

            Hard-spinning  Nanosleep
real       0.733s               0.763s
user      0.732s               0.312s  
sys        0.410s               0.407s

src/tool/hpcrun/gpu/gpu-trace.c
```diff
@@ -88,7 +89,7 @@
 
 #define DEBUG 0
 #include "gpu-print.h"
-
+#define TRACE_CHANNEL_SLEEP 5000000 // in ns
```
Jokeren, line 92: Isn't it 5ms instead of 5us?

Jokeren (changes requested):

jmellorcrummey (commented):
I am accepting the changes but I am changing the sleep time to half a second 50M ns. I don't think that we need to be checking for updates 200  times per second. 20 times a second should be plenty. Even delaying a half second might be fine.

src/tool/hpcrun/gpu/gpu-trace.c
dejangrubisic: Yes it is 5ms, my fault

PR #373, opened 2021-01-21
Minor edits to the GPU section of the manual

mxz297:


doc/manual/gpu.tex
```diff
@@ -69,6 +69,7 @@ \subsection{Performance Measurement of CUDA Programs}
   GXCOPY (sec)  &  GPU time: explicit data copy (seconds)  \\\hline 
 % GICOPY (sec)  &  GPU time: implicit data copy (seconds)  \\\hline 
   GSYNC (sec)  &  GPU time: synchronization (seconds)  \\\hline 
+ GPUOP (sec) & Total GPU operation time: the summation of all metrics above \\\hline
```
jmellorcrummey, line 72: summation --> sum

doc/manual/gpu.tex
```diff
@@ -322,7 +323,7 @@ \subsection{Attributing Measurements to Source Code for NVIDIA GPUs}
 When HPCToolkit uses NVIDIA's CUPTI to monitor a GPU-accelerated application, 
 CUPTI notifies HPCToolkit every time it loads a CUDA binary, known as a CUBIN, into a GPU.
 At runtime, HPCToolkit computes a cryptographic hash of a CUBIN's contents and records the CUBIN into the execution's measurement directory. 
-For instance, if a GPU-accelerated application loaded CUBIN into a GPU, NVIDIA's CUPTI informed HPCToolkit that the CUBIN was being loaded, and HPCToolkit computed its cryptographic hash as {\tt 972349aed8}, then HPCToolkit would record {\tt 972349aed8.cubin} inside a {\tt cubins} subdirectory of an HPCToolkit measurement directory. 
+For instance, if a GPU-accelerated application loaded CUBIN into a GPU, NVIDIA's CUPTI informed HPCToolkit that the CUBIN was being loaded, and HPCToolkit computed its cryptographic hash as {\tt 972349aed8}, then HPCToolkit would record {\tt 972349aed8.cubin} inside a {\tt gpubins} subdirectory of an HPCToolkit measurement directory.
```
jmellorcrummey, line 326: the cubin will actually be recorded with the suffix .gpubin.

This brings up a good point. We should probably reorganize this chapter of the manual into an overview section that describes coarse and fine-grain measurements.  We now collect gpubin objects for all three types of GPUs.  

Do we need to run hpcstruct on the AMD gpu binaries? We MUST run them on Intel GPU binaries if we use GT-Pin or we get nothing.

doc/manual/gpu.tex
```diff
@@ -384,7 +385,7 @@ \subsection{GPU Calling Context Tree Reconstruction}
 If no call instructions for a GPU function G have been sampled, the costs of G are apportioned evenly among each of G's call sites.
 \end{itemize}
 
-IHPCToolkit's {\tt hpcprof} analyzes the static call graph associated with each GPU kernel invocation. If the static call graph for the GPU kernel contains cycles, which arise from recursive or mutually-recursive calls,  \hpcprof{} replaces each cycle with a strongly connected component (SCC). In this case, \hpcprof{} unlink's call graph edges between vertices within the SCC and adds an SCC vertex to enclose the set of vertices in each SCC. The rest of \hpcprof{}'s analysis 
+IHPCToolkit's {\tt hpcprof} analyzes the static call graph associated with each GPU kernel invocation. If the static call graph for the GPU kernel contains cycles, which arise from recursive or mutually-recursive calls,  \hpcprof{} replaces each cycle with a strongly connected component (SCC). In this case, \hpcprof{} unlinks call graph edges between vertices within the SCC and adds an SCC vertex to enclose the set of vertices in each SCC. The rest of \hpcprof{}'s analysis
```
jmellorcrummey, line 388: FYI: This capability is nvidia specific now. it should not go into the general overview. Or, we could put it in the general overview, but note that it only works for NVIDIA to date.

jmellorcrummey (changes requested):

doc/manual/gpu.tex
mxz297: @jmellorcrummey 

> the cubin will actually be recorded with the suffix .gpubin.

Yes, I only change the directory name. 

> 
> This brings up a good point. We should probably reorganize this chapter of the manual into an overview section that describes coarse and fine-grain measurements. We now collect gpubin objects for all three types of GPUs.

Yes, I agree that such reogranization would be better. Right now, most information is under the CUDA subsection. 

> 
> Do we need to run hpcstruct on the AMD gpu binaries? We MUST run them on Intel GPU binaries if we use GT-Pin or we get nothing.

No, we do not need to run hpcstruct on AMD gpu binaries as AMD GPU binaries have standard DWARF for hpcprof to extract line info.

doc/manual/gpu.tex
mxz297: @jmellorcrummey This section is currently under section "NVIDIA GPUs".  The manual currently does not have a real general overview section for GPUs.

mxz297:
@jmellorcrummey I rebased upon your latest manual changes (description for partial unwind). This PR now reorganizes the GPU section and adds a paragraph about the `CUPTI_ERROR_NOT_INITIALIZED` error in the trouble shooting section.

jmellorcrummey (approved):

PR #374, opened 2021-01-25
Fixes regarding measurement directory

mxz297:
1. The rocm support code will directly return when finding out that the GPU binary exists, which will skip the steps to fill in data structures representing the list of GPU binaries. The fix is to just skip writing to disk but not the steps of filling the needed data structures.

2. Currently, running `hpcrun -L` will create an unwanted measurement directory. A full solution requires more thoughts on the ordering of the startup sequences. This PR includes a temporary solution that simply skips operations related creating measurement directory.


mxz297:
@mwkrentel Should work as a workaround for #377

mxz297:
@jmellorcrummey Update the branch to handle IO errors when writing AMD GPU binaries to disk and use `HPCRUN_LIST_EVENT` for `hpcrun -L` and put related code into one if conditional.

jmellorcrummey (approved):
LGTM.

PR #376, opened 2021-01-27
Fix PLT optimization for dlopen'd libraries

blue42u:
Test code:
```
#--- Makefile
all: main libfoo.so libbar.so
clean:
	rm -f main libfoo.so libbar.so

libbar.so: bar.c
	gcc -o $@ $^ -shared
libfoo.so: foo.c libbar.so
	gcc -o $@ -shared -Wl,-rpath=`realpath .` -L. $< -lbar
main: main.c libbar.so
	gcc -o $@ -Wl,-rpath=`realpath .` -L. $< -lbar -ldl

//--- main.c
#include <dlfcn.h>
extern void bar();
int main() {
  bar();
  void* h = dlopen("libfoo.so", RTLD_LAZY);
  void (*foo)() = dlsym(h, "foo");
  foo();
  dlclose(h);
}

//--- foo.c
extern void bar();
void foo() {
  bar();
  for(int i = 0; i < 1<<28; i++) bar();
}

//--- bar.c
void bar() {}
```

Output:
```console
$ \time ./main
0.46user 0.00system 0:00.46elapsed 99%CPU (0avgtext+0avgdata 1188maxresident)k
0inputs+0outputs (0major+87minor)pagefaults 0swaps
$ \time hpcrun -e REALTIME@0 ./main  # Before
22.56user 0.02system 0:22.62elapsed 99%CPU (0avgtext+0avgdata 11192maxresident)k
0inputs+144outputs (0major+12466minor)pagefaults 0swaps
$ \time hpcrun -e REALTIME@0 ./main  # After
0.50user 0.01system 0:00.53elapsed 96%CPU (0avgtext+0avgdata 9168maxresident)k
0inputs+96outputs (0major+12468minor)pagefaults 0swaps
```

### DIFF
--- a/src/tool/hpcrun/audit/audit-api.h
+++ b/src/tool/hpcrun/audit/audit-api.h
@@ -79,6 +79,9 @@ typedef struct auditor_map_entry_t {
 
   // Corrosponding load_module_t for this here thing.
   struct load_module_t* load_module;
+
+  // link_map entry for this library, if that makes sense for this entry.
+  struct link_map* map;
 } auditor_map_entry_t;
 
 typedef struct auditor_hooks_t {
--- a/src/tool/hpcrun/audit/auditor.c
+++ b/src/tool/hpcrun/audit/auditor.c
@@ -266,7 +266,7 @@ static void hook_open(uintptr_t* cookie, struct link_map* map, enum audit_open_f
   if(verbose)
     fprintf(stderr, "[audit] Delivering objopen for `%s' [%p, %p)"
 	    " dl_info.dlpi_addr = %p\n", entry->path, entry->start, 
-	    entry->end, entry->dl_info.dlpi_addr);
+	    entry->end, (void*)entry->dl_info.dlpi_addr);
 
   hooks.open(entry);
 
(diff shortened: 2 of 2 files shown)


jmellorcrummey:
Nice work! My comments notwithstanding, this addresses the problem we looked at earlier today.

jmellorcrummey:
FYI: I'm going to try out the symbind flag setting in my PLT unit test.

jmellorcrummey (approved):
FIrst, there are no comments in the code that describe why you changed from returning 0 to returning LA_FLG_BINDTO  | LA_FLG_BINDFROM.

While looking up information about these constants, I stumbled across the following:

If la_pltenter() or la_pltexit() routines exist, these routines are called after la_symbind() for procedure linkage table entries. These routines are called each time that the symbol is referenced. 

The following flags can be returned from la_symbind() to alter this default behavior. These flags are applied as a bitwise-inclusive OR with the value pointed to by the flags argument.

LA_SYMB_NOPLTENTER – Do not call the la_pltenter() routine for this symbol.

LA_SYMB_NOPLTEXIT – Do not call the la_pltexit() routine for this symbol.

By using these flags, would we just intercept the symbind for each symbol and then never incur any costs related to PLT calls thereafter? Or, do these flags mean that we execute the _dl_runtime_profile code and just skip around the PLT callbacks?

If we can skip the complicated patching and just return these constants from symbind, that might be best.

 **NOTE: I TESTED THE ABOVE IDEA AND IT IS AS SLOW AS AN EMPTY AUDITOR. YOUR IMPLEMENTATION IS THE WAY TO GO. I'M GOING TO MERGE.**

blue42u:
The relevant bit is at the bottom of the documentation for `la_objopen`, without them `la_symbind*` is never called:
```
       As its return value, la_objopen() returns a bit mask created  by  ORing
       zero  or  more of the following constants, which allow the auditing li‐
       brary to select the objects to be monitored by la_symbind*():

       LA_FLG_BINDTO    Audit symbol bindings to this object.

       LA_FLG_BINDFROM  Audit symbol bindings from this object.

       A return value of 0 from la_objopen() indicates that no symbol bindings
       should be audited for this object.
```
Since we don't expose any `la_pltenter` functions, the `LA_SYMB_NOPLT*` flags are automatically implied (`la_pltexit` is only called if `la_pltenter` is present. Its not documented.)

As it happens `la_symbind` is called regardless of the resolver, which means we do incur some additional cost to look up the GOTPLT table (thus the additional silent early exit in `optimize_object_plt`). The alternative would be to use `la_pltenter` (which doesn't get called after replacing the resolver), but I wanted a quick patch that didn't require N arch-specific signatures with `#ifdef` guards, so I went with this for now.

If it works well on the problematic case I'll clean it up a little and add some additional comments.

mxz297:
@blue42u Just out of curiosity. Do you know whether `la_symbind` is called before or after calling the actual resolver? 

The man page says that 

```
The return value of la_symbind32() and la_symbind64() is the
       address to which control should be passed after the function
       returns.  If the auditing library is simply monitoring symbol
       bindings, then it should return sym->st_value. 
```

This looks to me then during `la_symbind`, the symbol is already bound (otherwise, `sym->st_value` would not have the actual PLT call target), and therefore `la_symbind` seems to be called after calling the actual resolver.

If this is the case, then it seems to me that we will execute `dl_runtime_profile` once, and may trigger register corruption on ARM. However, I tested your code on ARM, there is no crash using our previous unit test. So I guess this is not an issue.

Do you know what I am missing here?

blue42u:
Its called after, within the bounds of the `_dl_runtime_*` call (although I don't remember the particulars off the top of my head). The actual resolver is called first to fill an `ElfW(Sym)` entry on the stack and then all the `la_symbind`'s are called before it gets written to the GOT (or shadow-GOT on the LD_AUDIT code path).

The idea is to let it call `_dl_runtime_profile` once per (applicable) binary and then set it to call `_dl_runtime_resolve` after that, `LA_CONSISTENT` is too early for `dlopen`'d libraries and `la_symbind` is the nearest next callback. This also makes sure it only gets called for binaries that actually have relocations.

I forgot about the ARM case, it shouldn't work in general (if the first PLT call in a library returns a struct it should crash). Our previous test had additional `printf`'s so removing those might make it break. The code given here (if `bar` is adjusted to return a large struct) should also crash under ARM.

PR #379, opened 2021-01-29
Gpu blocks

Jokeren:
Nothing special, just collect the number of block per GPU kernel instance to check if all sms are used.


jmellorcrummey (approved):
looks simple and useful.

PR #380, opened 2021-01-29
Fix the PLT optimizations (part 2)

blue42u:
Opens permissions on the entire GOT table instead of just the first few pages, since it appears that Glibc will lock it down early in the LD_AUDIT case (probably since LD_AUDIT uses a shadow-GOT).

This version uses the GNU_RELRO program header(s) to determine where to adjust, under the assumption that if the binary has none the linker won't know what to lock down (and so won't).

This also cleans up the program header handling, under the assumption that the binaries are always the same architecture as the auditor. This assumption prevails in the rest of the auditor code, if it turns out to be false (multi-arch application) we can adjust more completely then.

### DIFF
--- a/src/tool/hpcrun/audit/auditor.c
+++ b/src/tool/hpcrun/audit/auditor.c
@@ -121,6 +121,11 @@ enum audit_open_flags {
 };
 
 
+struct phdrs_t {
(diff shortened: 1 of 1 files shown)


jmellorcrummey (approved):
LGTM

PR #382, opened 2021-01-29
Disable auditor on ARM by default

mxz297:
Due to a glibc bug on ARM, the auditor cannot be reliably used. Also add `--enable-auditor` option to hpcrun so that the user can override the default choice.

Tested on ARM and x86.


jmellorcrummey (approved):
LGTM

PR #384, opened 2021-02-19
Opencl blame shifting

aarontcopal2:
- [x] Check that Blame shift works for OpenCL
- [ ] Check that Blame shift works for Cuda
- [ ] Check that Blame shift works for Rocm


Jokeren:
Can you describe the main features of this pull request? e.g., What problems you would like to target? What solutions you adopted to solve the problems? Is there any preliminary result for any application?

aarontcopal2:
> Can you describe the main features of this pull request?

In this pull request, I have implemented blame-shifting analysis for OpenCL. This is done after referring to Milind Chabbi's paper and his old branch in hpctoolkit on the same.

> What problems you would like to target?

With this pull request, the user can enable blame shifting by passing the CPU_GPU_IDLE event. e.g. `hpcrun -e gpu=opencl -e CPU_GPU_IDLE -e REALTIME ./application`
Primarily, we can see when both the CPU and GPU are idle. 3 metrics will be visible in hpcviewer: CPU_IDLE, GPU_IDLE and CPU_IDLE_CAUSE. 

> What solutions you adopted to solve the problems? 

I have created a generic interface for blame-shifting. I have used OpenCL api wrappers at certain points that call sync_start, sync_end, kernel_start, kernel_end, queue_start and queue_end functions of the blame-shifting interface. 

Since the interface is generic, other languages/platforms can use this to enable blame-shifting for itself.

>  Is there any preliminary result for any application?

I have tried this for Pelec. But since it has a small execution time, the metric values related to blame-shifting is small. So I plan to run this with a larger example. Im trying to build amr_wind for testing this out, but not yet successful.

mxz297:
@aarontcopal2 
 
> Since the interface is generic, other languages/platforms can use this to enable blame-shifting for itself.

I will evaluate this new interface next week to see if I can use it for AMD GPUs. 

> I have tried this for Pelec. But since it has a small execution time, the metric values related to blame-shifting is small. So I plan to run this with a larger example. Im trying to build amr_wind for testing this out, but not yet successful.

For PeleC, I thought you can increase the number of iterations at command line?

aarontcopal2:
@mxz297 

Sure. You can try it out and let me know.

> For PeleC, I thought you can increase the number of iterations at command line?

`./PeleC3d.dpcpp.ex inputs_ex max_step=10` is the command I use to run the application. can max_step take any natural number?

mxz297:
@aarontcopal2  I think  you can increase max_step to a larger integer to let it run longer.

Jokeren:
> I have created a generic interface for blame-shifting.

I suppose currently only OpenCL implementation is available? It might need some extra work to support CUDA. Milind previously used cuda events to monitor GPU kernels, but now we switch to CUPTI. I'll review your code and check it out.

> the metric values related to blame-shifting is small.

You can measure relative metrics. For example, the ratio of time spent in synchronization versus the total amount of time.

src/tool/hpcrun/gpu/gpu-activity.h
```diff
@@ -325,6 +326,13 @@ typedef struct gpu_branch_t {
 } gpu_branch_t;
 
 
+typedef struct gpu_blame_shift_t {
+  double cpu_idle_time;
+  double gpu_idle_time;
+  double cpu_idle_cause_time;
```
Jokeren, line 332: What is `cpu_idle_cause_time` and why there's no `gpu_idle_cause_time`?

src/tool/hpcrun/gpu/opencl/opencl-api.c
```diff
@@ -1207,6 +1271,74 @@ hpcrun_clCreateBuffer
...
+	// clWaitForEvents can wait on multiple events(probably from different queues).
+	// We need a more complicated approach of finding the queues on which the CPU will wait
+	// For now we pass the 1st queue
+	void *queue_ptr;
+	cl_command_queue queue;
+
+	if(is_opencl_blame_shifting_enabled()) {
+		size_t queue_size;
+		clGetEventInfo(*event_list, CL_EVENT_COMMAND_QUEUE, 0, NULL, &queue_size);
+		queue_ptr = hpcrun_malloc(sizeof(queue_size));
```
Jokeren, line 1292: Where this `queue_ptr` is recycled?

src/tool/hpcrun/gpu/gpu-activity.h
aarontcopal2: **how is cpu_idleness detected?**
call to sync() makes the cpu thread wait. In this case we know who is waiting(cpu thread) and who is the cause for wait(either the gpu stream or a kernel inside the stream). Thus we can increment both cpu_idle time and cpu_idle_cause appropriately.

**How is gpu idleness detected?** 
itimer callbacks check the number of active kernels periodically. When (#kernels == 0), the gpu_idle_time is incremented. Note that the stacktrace where the metric is recorded is responsible for the gpu idleness. gpu_idle_cause_time does not make sense. Where would we attribute that metric? And how will it be calculated?

src/tool/hpcrun/sample-sources/gpu/gpu-blame/active-kernels-map.c
```diff
@@ -0,0 +1,197 @@
...
+}
+
+
+void
+active_kernels_delete
+(
+ uint64_t ak_id
+)
+{
+  spinlock_lock(&ak_map_lock);
```
Jokeren, line 147: If this function is called by `clear_recursive`, you will perform `lock` and `unlock` the number of times equal to the number of nodes in the map.

src/tool/hpcrun/sample-sources/ss-list.h
```diff
@@ -104,4 +104,5 @@ SAMPLE_SOURCE_DECL_MACRO(level0)
 SAMPLE_SOURCE_DECL_MACRO(opencl)
 #endif
 #endif
+SAMPLE_SOURCE_DECL_MACRO(cpu_gpu_idle)
```
Jokeren, line 107: Why not rename `gpu_blame` to `cpu_gpu_idle` to keep it more consistent.

src/tool/hpcrun/sample-sources/gpu/gpu-blame/blame.c
```diff
@@ -0,0 +1,484 @@
...
+add_kernel_to_completed_list
+(
+ kernel_node_t *kernel_node
+)
+{
+  while (true) {
+    kernel_node_t *current_head = atomic_load(&completed_kernel_list_head);
+    atomic_store(&kernel_node->next, current_head);
+    if (atomic_compare_exchange_strong(&completed_kernel_list_head, &current_head, kernel_node)) {
+      break;
```
Jokeren, line 202: The above code is confusing to me. 

If you just want to append a node atomically to a concurrent list's head, you can follow the simple pattern below. No need to use an additional atomic load for list head.

https://github.com/HPCToolkit/hpctoolkit/blob/d53f6fce018673bc886305330ca91b686844468e/src/lib/prof-lean/stacks.c#L230

src/tool/hpcrun/gpu/gpu-activity.h
Jokeren: By reviewing Milind's paper, I reminded what `cpu_idle_cause` means. In fact, the original paper denotes `gpu_idle_time` as `gpu_idle_cause_time`.

Jokeren:
@aarontcopal2 Overall this pull request looks good, but needs very close attention to the concurrent data structures you are extensively using. I will take a second round look when I try to extend it for CUDA. 
One direction to further explore is to implement the full history blaming feature mentioned in Figure 12 of Milind's paper.

src/tool/hpcrun/gpu/gpu-activity.h
aarontcopal2: done

src/tool/hpcrun/sample-sources/gpu/gpu-blame/blame.c
aarontcopal2: @Jokeren , if the CAS in the while loop evaluates to false, where is the head variable getting updated?

aarontcopal2:
Thanks for the comments @Jokeren.

src/tool/hpcrun/gpu/opencl/opencl-api.c
aarontcopal2: the queue_size is variable. A simple alloc and free queue won't work. how is dynamic-sized hpcrun_malloc freed in hpcrun?

src/tool/hpcrun/sample-sources/gpu/gpu-blame/blame.c
Jokeren: The head variable is updated by `atomic_compare_exchange_strong`

src/tool/hpcrun/gpu/opencl/opencl-api.c
Jokeren: If it is in an application callback, just use malloc.
If only the current scope uses this variable, just declare it as a stack variable.

PR #385, opened 2021-02-22
ROCm support fix and GPU tracing fix

mxz297:
1. Do not record time stamps of HIP API executed on the host side. Record the time stamps of actual GPU operations from the GPU.
2. Fix GPU tracing, which requires matched time stamps for GPU operations and "no activity". This is a problem for all GPU platforms.


jmellorcrummey (approved):
Looks fine. 

To clean this up in the future, we should remove the unused

 uint64_t gpu_trace_prev_time;

field and figure out how to get the op codes from the function that turns the ops into the strings we are using.

PR #388, opened 2021-03-04
Monitor and tracing threads are not tracked by libmonitor and they ar…

dejangrubisic:
I added monitor_enable/disable_new_threads for each of the tracing threads and OpenCL monitor thread, together with pool allocation that we talked about in the meeting. Could you please check if this looks good. It compiles and generates expected DB for cu_multi_context_multi_stream example.

### DIFF
--- a/src/tool/hpcrun/gpu/gpu-trace.c
+++ b/src/tool/hpcrun/gpu/gpu-trace.c
@@ -358,6 +358,9 @@ gpu_trace_record
 {
   gpu_trace_channel_set_t *channel_set = (gpu_trace_channel_set_t *) args;
 
+  hpcrun_thread_init_mem_pool_once();
+  atomic_fetch_add(&active_streams_counter, 1);
+
   while (!atomic_load(&stop_trace_flag)) {
     //getting data from a trace channel
     gpu_trace_channel_set_process(channel_set);
@@ -378,14 +381,7 @@ gpu_trace_create
 {
   // Init variables
   gpu_trace_t *trace = gpu_trace_alloc();
-
-  // Create a new thread for the stream without libmonitor watching
-  monitor_disable_new_threads();
-
   trace->thread = gpu_trace_demultiplexer_push(trace->trace_channel);
-  atomic_fetch_add(&active_streams_counter, 1);
-
-  monitor_enable_new_threads();
 
   return trace;
 }
--- a/src/tool/hpcrun/thread_data.c
+++ b/src/tool/hpcrun/thread_data.c
@@ -200,6 +200,31 @@ hpcrun_threaded_data(void)
 }
 
 
+void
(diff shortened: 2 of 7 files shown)


src/tool/hpcrun/thread_data.c
```diff
@@ -200,6 +200,31 @@ hpcrun_threaded_data(void)
 }
 
 
+void
+hpcrun_thread_init_mem_pool_once(void){
+  static bool is_initialized = false;
```
Jokeren, line 205: I'm confused.

If the function is called by multiple threads, why isn't `is_initialized` a `pthread_once` variable. 

If `is_initialized` is supposed to be used individually by each thread, shouldn't `is_initialized` be a thread local variable?

src/tool/hpcrun/gpu/gpu-trace.c
```diff
@@ -378,14 +381,7 @@ gpu_trace_create
 {
   // Init variables
   gpu_trace_t *trace = gpu_trace_alloc();
-
-  // Create a new thread for the stream without libmonitor watching
-  monitor_disable_new_threads();
-
   trace->thread = gpu_trace_demultiplexer_push(trace->trace_channel);
-  atomic_fetch_add(&active_streams_counter, 1);
```
Jokeren, line 386: You may want to keep `atomic_fetch_add` here instead of the thread that you created.

Though unlikely, it is still possible that by the time the main thread reaches the end, the thread you created is suspended and never executed till `atomic_fetch_add`. In this way the main thread will see `active_streams_counter` as `actual streams - 1`.

src/tool/hpcrun/thread_data.c
dejangrubisic: yes you are right, I will change that

src/tool/hpcrun/gpu/gpu-trace.c
dejangrubisic: I did that because it happened that the tracing thread was killed before it started and we had a deadlock because the trace counter is never decremented.

Jokeren: How it is killed? I think the tracing thread is only created if there's any trace item. Then the tracing thread should have seen the counter value before you kill it.

dejangrubisic: It happened that libmonitor intercepted pthread_create (even though monitor_disable_new_threads was called before pthread_create) and there called real_pthread_attr_destroy so the thread was never really created (I posted this problem in hpcrun on Slack).  I asked Mark, but he is not sure why that happened. 

As result, I taught it is a good idea to increment and decrement stream counter from the same thread.

dejangrubisic:
I have just checked all files here and it looks to me that this PR is surpassed by patch_memory_pool_create

PR #389, opened 2021-03-10
Fixing hpcstruct for Intel GPU binaries when not specifying --gpucfg=yes

mxz297:


src/lib/banal/gpu/ReadIntelCFG.cpp
```diff
@@ -231,10 +231,13 @@ readIntelCFG
  Dyninst::ParseAPI::CodeObject **code_obj
 )
 {
-  if (cfg_wanted) {
-    auto function_name = elfFile->getGPUKernelName();
-    addCustomFunctionObject(function_name, the_symtab); //adds a dummy function object
+  // We need to create a dummy function object whether
```
jmellorcrummey, line 234: I think the comment could be improved for consumption by others who don't understand the details.

// An Intel GPU binary for a kernel does not contain a function symbol for the kernel 
// in its symbol table. Without a function symbol in the symbol table, Dyninst will not 
// associate line map entries with addresses in the kernel. To cope with this defect of 
// binaries for Intel GPU kernels, we add a function symbol for the kernel to its Dyninst 
// symbol table.

jmellorcrummey (changes requested):
I think the comment could be improved so we can understand this in the future. See my suggested text.

PR #390, opened 2021-03-13
GPU CFG construction fix for dyninst@10.2

Jokeren:
Use arch flag to determine the entry id of a GPU instruction; If using dyninst 10.2, intel GPU instruction is not supported


jmellorcrummey (approved):
LGTM

PR #391, opened 2021-03-14
Quick fix for fall through edge; needs more testing

Jokeren:
TODO LIST:

- [x] 1. Add support for call instructions.
- [x] 2. Dump GPU binaries of AMR-Wind and PeleC for testing.
- [x] 3. Test if dotgraph does not fail on CPU binaries and CUDA binaries.


mwkrentel:
@Jokeren  Could you fix the indentation in src/tool/hpcstruct/Makefile.am ?
Thanks

Jokeren:
@mwkrentel done, thanks for checking!

@mxz297 could you please review the code?

mxz297 (approved):
LGTM

src/lib/banal/Struct.cpp
```diff
@@ -1638,7 +1638,14 @@ doFunctionList(WorkEnv & env, FileInfo * finfo, GroupInfo * ginfo, bool fullGaps
...
 
-  if (cuda_arch == 0) {
+  // computeGaps relies Block::start() and Block::end() to have
+  // the correct address range for a block to detect gaps.
+  // We have subclass GPUBlock inheriting ParseAPI::Block.
+  // However, currently subclass of ParseAPI::Block cannot update
+  // block end, which leads to Block::end() == Block::start() for
+  // GPUBlock objects.
+  // Therefore, we cannot invoke gap computation for GPU binaries
+  if (cuda_arch == 0 && intel_gpu_arch == 0) {
```
jmellorcrummey, line 1648: isn’t it the case that blocks have a proper block end address and and meet the requirements for gap parsing?  
if we don’t want to do gap parsing for gpu binaries, that’s a
choice. however are the comments above that say we can’t still right?

jmellorcrummey (commented):
@Jokeren @mwkrentel @mxz297 i’d like an answer about the gap parsing comments before appeovi g the merge.

src/lib/banal/Struct.cpp
mxz297: @jmellorcrummey You are correct that the comments are not long valid. I am going to update the comment. I don't think we should invoke gap computation for GPU code even if we have presumably correct  block boundary since we have never tested its impact.

mxz297: @jmellorcrummey Comments updated.

PR #392, opened 2021-03-20
Fix kernel serialization problem cause by CUPTI_ACTIVITY_KIND_KERNEL;…

Jokeren:
… Using concurrent kernel monitoring brings overhead for GPU kernels with many short-lived blocks. Users can use SERIALIZATION control knob to switch between the concurrent and the serialization measurement modes, depending on the number of streams used by each application

### DIFF
--- a/src/lib/prof/NameMappings.cpp
+++ b/src/lib/prof/NameMappings.cpp
@@ -47,6 +47,7 @@ const char *GPU_COPYIN  = "<gpu copyin>";
 const char *GPU_COPYOUT = "<gpu copyout>";
 const char *GPU_ALLOC   = "<gpu alloc>";
 const char *GPU_DELETE  = "<gpu delete>";
+const char *GPU_MEMSET  = "<gpu memset>";
 const char *GPU_SYNC    = "<gpu sync>";
 const char *GPU_KERNEL  = "<gpu kernel>";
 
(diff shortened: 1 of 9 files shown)


Jokeren:
@dejangrubisic , please use this branch to profile PeleC, Quicksilver, and Laghos, reporting their corresponding kernel time under the two different modes.

These examples can be quickly built with the scripts here:

https://github.com/HPCToolkit/hpctoolkit-tutorial-examples/tree/nersc/examples/gpu

Jokeren:
You can also check the differences between two modes in the trace view.

src/lib/prof/NameMappings.cpp
```diff
@@ -47,6 +47,7 @@ const char *GPU_COPYIN  = "<gpu copyin>";
 const char *GPU_COPYOUT = "<gpu copyout>";
 const char *GPU_ALLOC   = "<gpu alloc>";
 const char *GPU_DELETE  = "<gpu delete>";
+const char *GPU_MEMSET  = "<gpu memset>";
```
Jokeren, line 50: @jmellorcrummey This is a top priority fix. Without adding the missing tag, hpcrun generates some weird call paths on summit. Then, hpcprof is confused and failed to identify the GPU kernel launch site. As a result, no GPU calling context tree will be reconstructed. 

This fix solves the above problem.

Jokeren: Update...

This does not seem to be the only culprit. There are other problems I am still investigating.

dejangrubisic:
@jmellorcrummey @Jokeren here are results for Laghos, PeleC and QuickSilver for -ck HPCRUN_CUDA_KERNEL_SERIALIZATION

https://docs.google.com/spreadsheets/d/1hGR6-Ozy_Abr-7CUXpKoYAB2LtfDIIgN2fp4VKDyJJ8/edit?usp=sharing

1st - hpcrun -e cycles -e gpu=nvidia -t
2nd - hpcrun -ck HPCRUN_CUDA_KERNEL_SERIALIZATION=TRUE -e cycles -e gpu=nvidia -t 

Laghos        0m7.578s     | 0m6.778s
PeleC           1m13.153s   | 1m22.223s
QuickSilver  0m3.627s     | 0m3.758s

Jokeren:
I checked the result with Dejan, and it looks reasonable to me.

In the PeleC case, the concurrent kernel mode has less overhead than the serialized mode because PeleC uses five GPU streams per process.

For the single stream Laghos, because the concurrent kernel mode instrument many small GPU kernels, the serialized mode without instrumentation can have less overhead.

jmellorcrummey (approved):
LGTM

PR #397, opened 2021-04-03
Fixed issues of uninitialized gpu_trace_finished and control_knob.c circular list

dejangrubisic:
The bug was created when we want to register a control knob that is already registered. Instead of just changing the value, I was mistakenly setting next pointers, which should be done only when adding a control knob the first time in the list.

### DIFF
--- a/src/tool/hpcrun/control-knob.c
+++ b/src/tool/hpcrun/control-knob.c
@@ -40,7 +40,6 @@ control_knob_register(char *name, char *value, control_knob_type type)
     iter->name = strdup(name);
     iter->type = type;
     iter->value = strdup(value);
-
     iter->next = control_knobs;
     control_knobs = iter;
   }else{
--- a/src/tool/hpcrun/control-knob.c
+++ b/src/tool/hpcrun/control-knob.c
@@ -39,10 +39,14 @@ control_knob_register(char *name, char *value, control_knob_type type)
     iter = (control_knob_t*) malloc(sizeof(control_knob_t));
     iter->name = strdup(name);
     iter->type = type;
+    iter->value = strdup(value);
+
+    iter->next = control_knobs;
+    control_knobs = iter;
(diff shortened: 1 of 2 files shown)


src/tool/hpcrun/control-knob.c
```diff
@@ -39,10 +39,13 @@ control_knob_register(char *name, char *value, control_knob_type type)
     iter = (control_knob_t*) malloc(sizeof(control_knob_t));
     iter->name = strdup(name);
     iter->type = type;
+    iter->value = strdup(value);
```
Jokeren, line 42: It can be simplified. Both branches execute `iter->value = strdup(value);`

Jokeren (changes requested):

Jokeren:
I am not sure which branch to merge, maybe develop is a better choice according to @mwkrentel .

jmellorcrummey:
I think this should be merged into master as it is a bug fix for the release rather than a complex new feature.

src/tool/hpcrun/control-knob.c
dejangrubisic: I was planning to do that, but my bug arose because of me trying to pull things out of if().
This way I think it is easier to read, and all functionality is self-contained in if/else.

I can change it if you want.

mwkrentel:
Yes, bug fixes should go to whatever branch has the bug.

I was suggesting that new development, especially for Intel should
start in a branch.

src/tool/hpcrun/control-knob.c
jmellorcrummey: please put the assignment iter->value=strdup(value) after the if. eliminate the assignment in the if branch and remove the else branch entirely.

jmellorcrummey:
I don't want to wait for dejan to make the recommended change. I will make it in master.

PR #405, opened 2021-05-07
draft of support for aborting a hung CUPTI flush

jmellorcrummey:


src/tool/hpcrun/gpu/nvidia/cupti-api.c
```diff
@@ -1513,14 +1599,23 @@ cupti_pc_sampling_disable
...
 void
 cupti_activity_flush
 (
 )
 {
   if (cupti_stop_flag) {
     cupti_stop_flag_unset();
-    HPCRUN_CUPTI_CALL(cuptiActivityFlushAll, (CUPTI_ACTIVITY_FLAG_FLUSH_FORCED));
+    FLUSH_ALARM_INIT();
+    if (!FLUSH_ALARM_FIRED()) {
```
Jokeren, line 1611: Why using this macro that always returns 0? Is it for debugging?

src/tool/hpcrun/utilities/linuxtimer.c
```diff
@@ -0,0 +1,177 @@
...
+}
+
+
+int
+linuxtimer_newsignal
+(
+ void
+)
+{
+  static __thread int next_signal = 4;
```
Jokeren, line 129: What does *4* mean here?

src/tool/hpcrun/gpu/nvidia/cupti-api.c
jmellorcrummey: There are two versions of this macro. When CUPTI_FLUSH_HANG_WORKAROUND is non-zero, the macro performs a setjmp (that returns 0 on success and 1 if returned to by longjmp). If CUPTI_FLUSH_HANG_WORKAROUND is 0, then FLUSH_ALARM_FIRED is alway 0 so that you execute the cuptiActivityFlushAll.

Jokeren: Thanks for the explanation. 

If I understand it correct, if `cuptiActivityFlushAll` does not return when time is up, we jump back to `FLUSH_ALARM_FIRED()`.

Now that it is returned by longjmp, `FLUSH_ALARM_FIRED()=0` so that we terminate execution of this thread without waiting for `cuptiActivityFlushAll`.

src/tool/hpcrun/utilities/linuxtimer.c
```diff
@@ -0,0 +1,177 @@
...
+  int ret = timer_create(clock, &td->sigev, &td->timerid);
+
+  if (ret) td->timerid = NULL;
+
+  return ret;
+}
+
+
+int
+linuxtimer_newsignal
```
mxz297, line 124: @jmellorcrummey Do we want to query whether the next_signal has been used by the application and skip the ones that have been registered?

PR #414, opened 2021-05-26
Workaround an issue with libmemkind.

mwkrentel:
Add two options:

 1. configure --with-memkind=PATH ...
 This compiles in the path to the libmemkind.so directory.

 2. hpcrun --memkind ...
 This activates adding the dir to LD_LIBRARY_PATH.

----------

Lightly tested, as in, I don't think it will break anything, but I
don't know if it will fix the original problem.  I'm unsure about two
things:

 1. The order of LD_LIBRARY_PATH.  If the app or openmpi has
 LD_LIBRARY_PATH, do we need to be on the front or back to override
 the app.

 2. If openmpi rpath's their directory, I don't know that LD_LIBRARY_PATH
 will have any affect.

ping @jmellorcrummey for more advice on testing in battlefield conditions.


mwkrentel:
That's curious, I merged this manually from the command line and 
github immediately noticed and declared the PR closed and merged !

PR #416, opened 2021-07-09
STREAMS_PER_TRACING_THREAD default set to 256 and gpu_trace_cct_no_activity optimization

dejangrubisic:
Based on the latest experiment we(John and I) conclude that one tracing thread is capable of dealing with 256 GPU streams.

Based on Oracle PE, we put gpu_trace_cct_no_activity out of the trace-consuming loop and get some performance.

### DIFF
--- a/src/tool/hpcrun/gpu/gpu-trace-channel.c
+++ b/src/tool/hpcrun/gpu/gpu-trace-channel.c
@@ -225,6 +225,7 @@ gpu_trace_channel_consume
   // reverse them so that they are in FIFO order
   channel_reverse(channel, bichannel_direction_forward);
 
+  cct_node_t *no_activity = gpu_trace_cct_no_activity(channel->td);
   // consume all elements enqueued before this function was called
   for (;;) {
     gpu_trace_item_t *ti = channel_pop(channel, bichannel_direction_forward);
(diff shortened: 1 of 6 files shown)


src/tool/hpcrun/gpu/gpu-trace-channel.c
```diff
@@ -236,7 +237,7 @@ gpu_trace_channel_consume
            ti->start,
            ti->end,
            ti->call_path_leaf);
-    gpu_trace_item_consume(consume_one_trace_item, channel->td, ti);
+    gpu_trace_item_consume(consume_one_trace_item, channel->td, ti, no_activity);
```
Jokeren, line 240: Good point.

It is a little bit weird though to pass a no activity cct all the way down to the place where it is being used. Besides, if gpu_trace_channel_consume is getting used multiple times, you still have to call gpu_trace_cct_no_activity multiple times. Is it possible to not change the interface but instead associate a `no_activity` item with each thread local data?

jmellorcrummey: The no_activity field was in thread-local data. We were paying a significant cost to fetch <no activity> from thread-local data once for each GPU operation to write <t, no activity> <t+epsilon, gpu kernel> <t+kernel length, gpu kernel>< t+kernel length+epsilon,no activity>. We use a loop to log many GPU operations for a stream at once. To amortize the cost of fetching <no activity> from thread-local data, we had to lift it out of gpu_trace_item_consume so that we could fetch it once for a run of (perhaps) many GPU operations on a stream rather than once for each operation.

jmellorcrummey (approved):

PR #417, opened 2021-07-15
Prof2 -> develop

blue42u:
This PR merges the current latest of Prof2 development into develop. The changes are broken coarsely into 5 commits, parts 2, 3 and 4 may be somewhat reviewable, part 1 is very much not and part 5 does not need it. I'm not expecting an in-depth code review, 28K lines of changes is nothing to sneeze at.

Although Prof2 is incomplete and still in development, it should be stable enough for use within the group. All analysis work moving forward should be (and has been) based on Prof2, to support the new features planned for the next release Prof2 needs to be user-ready by the next release cycle. In keeping with precedent (and against my better judgement), this means merging it into develop and updating the topic branches accordingly.

Note that Prof2 is designed to work on (almost) C++17 with GCC 7.x and up. It probably does not compile flawlessly on all of these versions yet, until this is resolved I highly recommend using GCC 9.x and up. I would also recommend all new code based on Prof2 adhere to these restrictions and use modern C++ when appropriate.

The remaining tasks for Prof2 are split between a Google Doc and my personal task list. If anyone is interested I can convert these to Github issues or a Kanban (Github "Projects" tab). Either would also help to keep track of the (definitely lingering) bugs that will be found as everyone updates.

With this PR **`hpcprof` will no longer work**, even though the code is still in the tree it will not function as intended. Same goes for `hpcproftt` on old measurements. The old-Prof code is thus safe to remove along with corresponding Autogoo, in particular AFAIK `libbfd` will no longer be a dependency since Prof2 uses Elfutils directly (so no more mysterious link errors when you forget to set `binutils~nls` in Spack).

There are also some major differences between how Prof2 and old-Prof match binaries + Structfiles + source files, notably `-I` is not implemented. Changes can be made if someone is willing to develop a proper design for this process and the command line interface, this may mesh well with @martyitz's usability interviews. To address in particular: finding source/binary candidate paths from `.hpcrun`/Structfiles/DWARF, breaking ties between equally valid candidates quickly, (weakly) verifying that said files match the application's state during measurement, deciding where and when to apply Structfiles, and proper concise warnings + fallbacks when "intention" from the command line does not match reality. All while considering renamed `.hpcrun`/source/binary files, relative paths from DWARF, and recompiled/updated binaries.

@Lucindaaaa If you're planning on working more with Prof2 please use this branch, this one is updated against `master`. In theory it should work just as well as `prof2` since the patches are nearly identical.

@laksono Currently HPCViewer is not able to read the generated database, I don't recall everything that has changed since it did (I remember the change to triples in the id-tuples). This should be fixed (at least to not crash) before this PR is merged.

@Jokeren @dejangrubisic The changes needed to support low-overhead NVIDIA sampling and trace sorting should be present in this PR, there are a few `#define`'s that should be replaced with `#include`'s as more gets merged in but that can be addressed later.

@jmellorcrummey @mwkrentel I will also update #320 to port the build system changes once this is merged, and once again if/when old-Prof is cleaned out of the tree.


blue42u:
Updates:
 - Errors on GCC < 9.x have been fixed (I think), now all GCC >= 7.x should work.
 - @laksono has updated the Viewer to support the latest format, which should work without issues now. Thanks!
OP edited to reflect these updates.

blue42u:
Update:
 - More fixes for the experiment.xml output (fixed missing \<LoadModule> tags)
 - New (probably) working support for kernel sampling (added support for pulling function data from files in `kernel_symbols`, currently always assumes liblzma compression)

blue42u:
Updates:
 - Remove the `trace.db` footer from `FORMATS.md` to match the current Prof2 output, and remove the corresponding check (and error message) from `hpcproftt`
 - Fix a bug with Prof2-MPI related to the hacks to handle `line` Scopes

PR #421, opened 2021-08-24
augment hpcstruct to analyze all cpu binaries in measurements directory

jmellorcrummey:
- refine parallel makefile so it can analyze cpu binaries as well as gpu binaries
- analyze both cpu and gpu binaries by default
- support parallel and concurrent analysis
- picks up where it left off if interrupted and restarted


jmellorcrummey:
So far, I have tested this on ufront. I will check this out and try it on a Power9 + nvidia.

src/tool/hpcstruct/pmake.txt
```diff
@@ -0,0 +1,198 @@
...
+# enable analysis of CPU binaries
+#*******************************************************************************
+ifeq ($(CPU_ANALYZE),1)
+CPUBIN_DIR  = $(MEAS_DIR)/cpubins
+LM_DIR      = $(MEAS_DIR)/lm
+
+#-------------------------------------------------------------------------------
+# $(H): hpcrun file for thread 0 in each process
+#-------------------------------------------------------------------------------
+H := $(wildcard $(MEAS_DIR)/*-000000-000-*.hpcrun)
```
Jokeren, line 59: @jmellorcrummey What if there's no thread 0 in this measurement dir or thread 0 does not contain all necessary load modules?

jmellorcrummey: When the measurement library is initialized, there is always a thread 0. It is possible (but unlikely) that thread 0 will finish early and only record a prefix of the roadmap. I decided to choose thread 0 rather than processing hpcrun files for all threads, which will typically be duplicates of thread 0. I think that this will do the right thing in all but exceptional cases. 

We could add an option to consider all threads.

Jokeren: Thanks for the explanation!

> We could add an option to consider all threads.

That makes sense.

jmellorcrummey: @Jokeren After further consideration, I decided to analyze all hpcrun files all the time. Scraping load modules from hpcrun files should be much faster than running hpcstruct on anything, so I don't think it will be a bottleneck. 

Other changes: 
- Don't remove files with listing of load modules per hpcrun file. this improves incrementality if it is interrupted and restarted. 
- Ensure that pushd output is suppressed.

jmellorcrummey:
adjusted to work on directories that lack gpubins

jmellorcrummey:
@mwkrentel @martyitz can you review this commit today? I'd like to merge it to master.

I will try to test it at Sandia today on Victor's application run.

mwkrentel:
yes, you're up next after a spack telecon starting now

PR #425, opened 2021-09-07
Fix naming for stripped functions

mxz297:
Functions from stripped binaries have functions names "outline" as a prefix and have redundant load module in their names. Such names are misleading. Use "unknown procedure @address [load module]" for their name.


jmellorcrummey (approved):
Looks good to me. We reviewed  the results of the patch together and they are a distinct improvement.

PR #426, opened 2021-09-08
Ignore GPU synchronization events as they are host side events

mxz297:


Jokeren (approved):

jmellorcrummey (approved):
This patch completely removes GPU synchronization from CUPTI monitoring. It doesn't help with GPU-side monitoring and on the host side, we already see samples in CUDA synchronization routines.

PR #429, opened 2021-09-19
Fix a couple of POSIX shell compatibility bugs in the hpcstruct Makefile

blue42u:
The new version of the Struct Makefile (for processing an entire measurements directory) do not work out-of-the-box on my Debian rig, since in Debian `/bin/sh` is not bash but a minimal POSIX shell called [dash](https://manpages.debian.org/unstable/dash/dash.1.en.html).

This patch removes the reliance on two bash-isms that aren't present in all POSIX-compliant shells:
 - In bash, `&> file` or `>& file` can be used to redirect both standard output and error to the same file, being equivalent to `> file 2>&1`. The Makefile only uses this to redirect to `/dev/null`, so the constructs are replaced with `>&- 2>&-` which closes standard output and error for the same effect.
 - Bash supports a `pushd` builtin to manipulate the "directory stack," a mechanism for saving working directories over the course of a session. The Makefile makes absolutely no use of the directory stack, so the (single) use of `pushd` is replaced with the equivalent `cd`.


jmellorcrummey (approved):
LGTM.

PR #441, opened 2021-09-24
Hotfix the grandchild Fnbound spawn (fixes bus error on ARM)

blue42u:
As per the email thread, the changes in #423 added a bug on ARM where the grandchild would cause a bus error. Although I don't have an exact rationale why this would occur, separating the child and grandchild stacks seems to resolve the issue.

### DIFF
--- a/src/tool/hpcrun/fnbounds/fnbounds_client.c
+++ b/src/tool/hpcrun/fnbounds/fnbounds_client.c
@@ -452,12 +452,9 @@ hpcfnbounds_child(void* fds_vp) {
     pid_t pid;
   }* fds = fds_vp;
 
-  // We should be roughly able to figure out which direction the stack grows.
-  void* stack = (void*)&fds < (void*)&server_stack[SERVER_STACK_SIZE * 1024]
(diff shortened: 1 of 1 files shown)


jmellorcrummey (approved):
What a subtle bug! The child and the grandchild were sharing the same memory for the stack. If the grandchild got started before the child finished, the grandchild could overwrite a child's return address on the stack.
Stack allocating space for the child stack in the parent process is fine because we wait for it to finish before returning from the parent stack frame where child_stack is allocated. I accept the pull request on the basis of inspection, but will test on theta anyway!

blue42u:
I'm not super satisfied with that explanation, the child shim exits without an error and the parent process continues along its merry way to send an ACK to the Fnbounds process. But somehow the grandchild shim gets a SIGBUS very early in its body (opening brace of function), so I would normally assume its the child putting the grandchild in a bad way. The only time the parent or child got a SIGBUS was when the grandchild was also sharing the address space.

Regardless, separating the stacks is easy and much cleaner than what I wrote before, so assuming it works I'm happy with this solution.

PR #443, opened 2021-09-25
Various fixes related to CPU tracing

mxz297:
1. Add <no activity> to CPU traces when there is a long gap without any sample
2. Only do cpu tracing when a time based metric is specified, such as cycles, cputime, and realtime.
3. Print an error message when -t is specified but no suitable tracing metric is given
4. fix itimer event's handling of sampling frequency
5. Cut GPU traces size by half


jmellorcrummey (approved):
Looks good to me. Tested on quicksilver on ufront.

I added a commit that adjusted some comments, a variable name, and flag names for clarity.

PR #444, opened 2021-09-26
Fix hpcviewer.e4 bug #108 (partial call path in flat view)00

laksono:
The cause of the bug is because hpcprof generated the 'f' attribute of <Partial call paths> to "1" (folder node) instead of "4" (top down folder).
By setting the value to "4" fixes this problem.


jmellorcrummey (approved):
this should fix the problem we observed with partial call paths in the flat view.

PR #445, opened 2021-09-27
overhaul hpcstruct to be parallel by default

jmellorcrummey:
- add parallelism by default:
    -- default concurrency: cpuset size /2
    -- default parallelism: min(16, cpuset size/2)
- have job control throttle parallelism
- error message when output file for directory
- rewrite usage message
- migrage usage message to usage.txt
- add path to hpcstruct to makefile
- bulletproof parallel makefile for exceptional cases
    -- no measurement files

### DIFF
--- a/src/lib/prof-lean/Makefile.am
+++ b/src/lib/prof-lean/Makefile.am
@@ -96,6 +96,7 @@ MYSOURCES = \
 	\
 	binarytree.h binarytree.c \
 	cskiplist.h cskiplist.c \
+	cpuset_hwthreads.h cpuset_hwthreads.c \
 	crypto-hash.h crypto-hash.c \
 	queues.h queues.c \
 	stacks.h stacks.c \
--- a/src/lib/prof-lean/Makefile.in
+++ b/src/lib/prof-lean/Makefile.in
@@ -151,6 +151,7 @@ am__objects_1 = libHPCprof_lean_la-hpcrun-fmt.lo \
 	lush/libHPCprof_lean_la-lush-support.lo \
 	libHPCprof_lean_la-binarytree.lo \
 	libHPCprof_lean_la-cskiplist.lo \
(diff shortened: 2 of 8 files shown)


martyitz:
I did not review the code in detail, but I have reviewed usage.txt.  I have the following suggestions:

Lines 2-4 — Shouldn’t it refer to the measurements directory before the usage for an individual binary?
Line 26 — For concurrent runs, is the default CPU-set/2, or is it the min of 16 and CPU-set/2?
Line 37 — Shouldn’t it say that -j can only decrease the parallelism, not increase it?
Line 43 — We can’t use STRINGIFY in the man page.
Line 53 — Isn’t the gpucfg default no for Nvidia, but yes for other architectures?
Line 68 — Wouldn’t it be better to have --time always enabled?

If you agree, I will make those changes after yor do the pull to master and I update the man page to match.

jmellorcrummey:
> Lines 2-4 — Shouldn’t it refer to the measurements directory before the usage for an individual binary?
Done.
> Line 26 — For concurrent runs, is the default CPU-set/2, or is it the min of 16 and CPU-set/2?
The default implemented is cpuset/2. The multithreaded default is min(cpuset/2,16).
> Line 37 — Shouldn’t it say that -j can only decrease the parallelism, not increase it?
The way it is implemented, it can increase it as well. I want that flexibility for people who want speed.
> Line 43 — We can’t use STRINGIFY in the man page.
Eliminated use of stringify.
> Line 53 — Isn’t the gpucfg default no for Nvidia, but yes for other architectures?
Right now, it is no for all. It is sort of moot for Intel and AMD at present.
> Line 68 — Wouldn’t it be better to have --time always enabled?
Not for this time, which times different phases of hpcstruct.
> 
> If you agree, I will make those changes after yor do the pull to master and I update the man page to match.

I made these changes. You can add them to the man page.

PR #447, opened 2021-09-27
Rework the rules for how struct chooses the file/line for a loop.

jmellorcrummey:
Don't avoid all inline steps as in 2b6dda9a1a.  Instead, use the
inline step if it containes an exit cond at a deeper level.

For one example analyzing hpcstruct, this reduces the number of loops
at line 0 from 77 to 4 (not counting PLT examples).

Add extra debugging output for the file and line choices.

### DIFF
--- a/src/lib/banal/Struct.cpp
+++ b/src/lib/banal/Struct.cpp
@@ -314,6 +314,7 @@ public:
   long  file_index;
   long  base_index;
   long  line_num;
+  vector <FLPIndex> flp_path;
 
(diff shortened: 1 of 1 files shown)


mwkrentel:
Ok, I did a few more tests: arm, static, intel.
In all cases, I'm getting comparable results, where the new heuristic brings
the remaining number of loops at line 0 down to the single digits (where line map
info is available).

So, I'm ready to merge.

PR #451, opened 2021-10-05
Restrict the dlsym("main") lookup to times when CHECK_MAIN is enabled.

mwkrentel:
There are reports that glibc on summit crashes for this and AFAICT,
the result is only used with CHECK_MAIN.

ping @jmellorcrummey  @blue42u

### DIFF
--- a/src/tool/hpcrun/main.c
+++ b/src/tool/hpcrun/main.c
@@ -284,10 +284,15 @@ setup_main_bounds_check(void* main_addr)
   // dynamic symbol named 'main' (if any).
   // passed into libc_start_main as real_main. this might be a
   // trampoline in the PLT.
-  dlerror();
-  main_addr_dl = dlsym(RTLD_NEXT,"main");
-  if (main_addr_dl) {
-    fnbounds_enclosing_addr(main_addr_dl, &main_lower_dl, &main_upper_dl, &lm);
+  //
+  // dlsym("main") fails on some broken glibc (early access summit),
+  // but we only use this when CHECK_MAIN is enabled.
+  if (ENABLED(CHECK_MAIN)) {
+    dlerror();
+    main_addr_dl = dlsym(RTLD_NEXT,"main");
+    if (main_addr_dl) {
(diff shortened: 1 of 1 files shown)


jmellorcrummey (changes requested):
I am less than satisfied with this patch. There are several broader issues.
1. the routine setup_main_boulds_check is passed an address known as main_addr. This address can only be the address of the main procedure in a static binary because that is all that monitor_get_main_address() returns. (I looked at the code.) Why pass main_addr into this routine. A better name for the libmonitor routine would be monitor_get_static_main(), but we're not going to change that today. How about moving the call to moinitor_get_main_address() into the setup_main_bounds_check and having it assign a variable static_binary_main. before a conditional that checks if(main_addr_static).
2. Later in this file, the debugging function hpcrun_inbounds_main checks the bounds against main_lower_dl and main_upper_dl. I think the code should check if (ENABLED(CHECK_MAIN)) there as well and print a warning if hpcrun_inbounds_main is called but CHECK_MAIN is not set so the _dl addresses haven't been set for a comparison. Another approach would be to call the setup routine if the _dl addresses haven't already been set. If any of the addresses main_lower, main_upper, main_lower_dl, main_upper_dl are zero, this routine should probably warn that they haven't been initialized.

mwkrentel:
@jmellorcrummey
Well, I guess I badly misunderstood my task for this.

> 1.  the routine setup_main_boulds_check is passed an address known as
> main_addr. This address can only be the address of the main procedure
> in a static binary because that is all that monitor_get_main_address() returns.

No, that's not correct.  Monitor's view of `main()` is the symbol
named "main" in the static case, or whatever function is passed to
`__libc_start_main()` in the dynamic case, and that's what
`monitor_get_addr_main()` returns.  It does cover the normal,
dynamic case.

The code in question here is a third case where "main" is not linked
directly into the main executable but loaded from shared library.
This is something you added in 2014.

I wrote the PR as a simple workaround to the glibc bug reported by
@blue42u and I still believe it does that.

If we what we need is a more extensive fix for the "third" case, then
we need a longer discussion about how that looks.

PR #454, opened 2021-10-12
Fix the need of -t to enable GPU CFG construction

mxz297:
With the help of @Jokeren , this fixes the bug that PC sampling requiring -t. I tested this PR with quicksilver on ufront. I compared the a few calling contexts with/without -t and the they look the same. The database with pc sampling without -t still have GPU call frames.

I am not certain about the side effect of this change related to CPU profiling.


jmellorcrummey (approved):
looks good to me. there is nothing to it!

PR #456, opened 2021-10-19
Avoid hanging at exit when using CLONE_VM on Glibc <2.24

### LINKED ISSUE #455: Hpcrun hangs at exit with older Glibc
Title says it all. Observed in `master`.

~~It appears the OpenMP threads are simply waiting around during the shootdown, the shootdown signal seems not to be delivered to these threads. Still investigating if the cause is an unknown modification to the signal disposition(s).~~

~~May have some interaction with #423.~~

**UPDATE:** I've figured out the issue, the OpenMP is really a red herring (easiest way to trigger the issue, but not actually the cause). In order:
1. The Fnbounds server is launched through a pair of calls to clone(). The clone() function is actually a Glibc wrapper around the clone syscall, the wrapper adds some code into the child to reset bits to make sure pthreads work properly. For obvious reasons its sensitive to whether the clone() is a "thread-like" or a "process-like".
2. Later, an OpenMP implementation spawns a thread team. Nothing special here, libmonitor tracks all the threads like it should.
3. When the application exits, libmonitor finalizes before OpenMP so the threads are still alive. This is a normal case, and causes libmonitor to decide to start the shootdown.
4. As part of the shootdown, libmonitor uses pthread_kill to send the shootdown signal to make the threads drain as they should. But the threads never get the signal.

Now for the chaotic chain of causes:
- The pthread_kill in (4) is actually throwing an error, but libmonitor pays no heed and stubbornly keeps trying without making any visible progress.
- The pthread_kill fails because its internal syscall to tgkill fails. tgkill takes both the pid and tid of the target (as a way to make sure the tid didn't get reused over process boundaries), in this case the pid is incorrectly -1.
- The reason the pid is -1 is because the Glibc internal pthreads state was reset, by the Glibc wrapper in the child process, through the shared CLONE_VM memory space, from (1).
- ...And this is a bug in Glibc, which was fixed in 2.24 (https://sourceware.org/bugzilla/show_bug.cgi?id=19957). NOTS and llnl.cs both run 2.17, and if my theory is sound this should also show up on gpu.cs.

Based on the bug reports a bunch of things go wrong when the pthread state is randomly reset. To be honest I'm surprised there weren't more mysterious crashes.

blue42u:
Fixes #455, ping @jmellorcrummey and @mwkrentel. IMO 2.24 is old enough we could get away with just dropping support, if not though this PR provides a hacky fix.

There is a bug in older Glibc (https://sourceware.org/bugzilla/show_bug.cgi?id=18862) where using `CLONE_VM` instead of `CLONE_THREAD` causes the Glibc wrapper to reset its internal pthread state in the parent process. This wreaks havoc with all things pthread, but most noticably it stops the `pthread_kill` in libmonitor's shootdown from functioning (and libmonitor doesn't pay attention enough to the returned error codes).

As a sort of hack, it turns out the reset happens after the body passed to `clone`, so killing the child via a signal instead of returning dodges the issue. At least, it seems to work on 2.17.

This also adds an auditor-export for `sigaction` (to reset the signal disposition for a faster signal-death). In the fake-auditor case there isn't a "default" function to use since `sigaction` is wrapped by libmonitor, which in turn doesn't provide a `monitor_real_sigaction`. Caveat emptor.

### DIFF
--- a/src/tool/hpcrun/fnbounds/fnbounds_client.c
+++ b/src/tool/hpcrun/fnbounds/fnbounds_client.c
@@ -386,7 +386,7 @@ shutdown_server(void)
   // collect the server's exit status to reduce zombies.  but we must
   // do it only for the fnbounds server, not any application child.
   if (server_pid > 0) {
-    auditor_exports->waitpid(server_pid, NULL, 0);
+    auditor_exports->waitpid(server_pid, NULL, __WCLONE);
   }
   server_pid = 0;
 
@@ -454,7 +454,7 @@ hpcfnbounds_child(void* fds_vp) {
 
   // Do the clone, and pass the result back to our parent through the shared memory space
   fds->pid = auditor_exports->clone(hpcfnbounds_grandchild,
-    &server_stack[SERVER_STACK_SIZE * 1024], SIGCHLD | CLONE_VM, fds_vp);
+    &server_stack[SERVER_STACK_SIZE * 1024], CLONE_UNTRACED | CLONE_VM, fds_vp);
   return 0;
 }
 
@@ -496,7 +496,7 @@ launch_server(void)
 
   // For safety, we don't assume the direction of stack growth
   child_pid = auditor_exports->clone(hpcfnbounds_child,
-    &child_stack[4 * 1024], SIGCHLD | CLONE_VM, &fds);
+    &child_stack[4 * 1024], CLONE_UNTRACED | CLONE_VM, &fds);
 
   if (child_pid < 0) {
     //
@@ -508,7 +508,7 @@ launch_server(void)
 
   // Wait for the child to successfully clone the grandchild
   int status;
-  if (auditor_exports->waitpid(child_pid, &status, 0) < 0) {
+  if (auditor_exports->waitpid(child_pid, &status, __WCLONE) < 0) {
     //
     // waitpid failed
     //
--- a/src/tool/hpcrun/fnbounds/fnbounds_client.c
+++ b/src/tool/hpcrun/fnbounds/fnbounds_client.c
@@ -545,7 +545,7 @@ launch_server(void)
   my_pid = getpid();
   client_status = SYSERV_ACTIVE;
 
-  TMSG(FNBOUNDS_CLIENT, "syserv launch: success, child shim: %d, server: ???", (int) child_pid;
+  TMSG(FNBOUNDS_CLIENT, "syserv launch: success, child shim: %d, server: ???", (int) child_pid);
 
   // Fnbounds talks first with a READY message
   struct syserv_mesg mesg;
--- a/src/tool/hpcrun/fnbounds/fnbounds_client.c
+++ b/src/tool/hpcrun/fnbounds/fnbounds_client.c
@@ -168,7 +168,6 @@ static int fdout = -1;
 static int fdin = -1;
 
 static pid_t my_pid;
-static pid_t server_pid = 0;
 
 #if 0
 // Limit on memory use at which we restart the server in Meg.
@@ -383,13 +382,6 @@ shutdown_server(void)
   fdin = -1;
   client_status = SYSERV_INACTIVE;
 
-  // collect the server's exit status to reduce zombies.  but we must
-  // do it only for the fnbounds server, not any application child.
-  if (server_pid > 0) {
-    auditor_exports->waitpid(server_pid, NULL, __WCLONE);
-  }
-  server_pid = 0;
-
   TMSG(FNBOUNDS_CLIENT, "syserv shutdown");
 }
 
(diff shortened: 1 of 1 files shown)


src/tool/hpcrun/fnbounds/fnbounds_client.c
```diff
@@ -455,7 +455,17 @@ hpcfnbounds_child(void* fds_vp) {
...
+
+  // Glibc before 2.24 will reset the internal pthread state if we return from
+  // this function. Since we're sharing a memory space with the parent this
+  // wreaks havoc. So instead die by a signal to get out.
+  struct sigaction act;
+  act.sa_handler = SIG_DFL;
+  act.sa_flags = 0;
+  auditor_exports->sigaction(SIGUSR1, &act, NULL);
+
+  raise(SIGUSR1);
```
jmellorcrummey, line 467: This idiom doesn't work on NOTS. The raise returns and then abort terminates with a SEGV. This causes everything to die, including the application. 
- I don't know why the process that raised an exception on itself doesn't die.
- I don't know why the child's death takes down the parent.

blue42u:
Simple reproducer for the GDB bug:
```console
$ cat x.c
#define _GNU_SOURCE
#include <sched.h>
#include <sys/wait.h>
#include <stdio.h>
#include <stdlib.h>
#include <time.h>
#include <unistd.h>

int body(void* _) {
  struct timespec ts = {2, 0};
  while(nanosleep(&ts, &ts));
  return 0;
}

int main() {
  void* stack = malloc(16*2*1024);
  int child = clone(body, stack+(16*1024), SIGCHLD | CLONE_VM, NULL);
  printf("Waiting for child: %d\n", child);
  waitpid(child, NULL, __WALL);
  printf("Child waited for\n");
  return 0;
}
$ gcc x.c -g
$ gdb ./a.out
...
(gdb) b x.c:17
Breakpoint 1 at 0x11da: file x.c, line 17.
(gdb) b x.c:19
Breakpoint 2 at 0x1221: file x.c, line 19.
(gdb) r
Starting program: /tmp/a.out 

Breakpoint 1, main () at x.c:17
17	  int child = clone(body, stack+(16*1024), SIGCHLD | CLONE_VM, NULL);
(gdb) c
Continuing.
[Detaching after fork from child process 298576]
Waiting for child: 298576
Child waited for
[Inferior 1 (process 298566) exited normally]
(gdb)
```

Weirdly enough forcing GDB to step slowly through the `clone` syscall makes it work again:
```console
(gdb) catch syscall clone
Catchpoint 3 (syscall 'clone' [56])
(gdb) r
Starting program: /tmp/a.out 

Breakpoint 1, main () at x.c:17
17	  int child = clone(body, stack+(16*1024), SIGCHLD | CLONE_VM, NULL);
(gdb) c
Continuing.

Catchpoint 3 (call to syscall clone), clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:78
78	../sysdeps/unix/sysv/linux/x86_64/clone.S: No such file or directory.
(gdb) c
Continuing.
[Detaching after fork from child process 298916]

Catchpoint 3 (returned from syscall clone), clone () at ../sysdeps/unix/sysv/linux/x86_64/clone.S:78
78	in ../sysdeps/unix/sysv/linux/x86_64/clone.S
(gdb) c
Continuing.
Waiting for child: 298916

Breakpoint 2, main () at x.c:19
19	  waitpid(child, NULL, __WALL);
(gdb) c
Continuing.
Child waited for
[Inferior 1 (process 298897) exited normally]
(gdb) 
```

blue42u:
Updated with the "backup plan," new patch description below. Rebased on https://github.com/HPCToolkit/hpctoolkit/pull/457 to prevent conflicts during the merge.

The fix here is to remove `CLONE_VM`, which as a side effect removes the parent's tracking for the Fnbounds process. Although there is a "shutdown" message that could be sent to close Fnbounds, since sending the message may block indefinitely this is not used to signal a shutdown (unchanged from the original). In the worst case scenario of an Fnbounds hung between queries, the application will exit without concern even though Fnbounds has not yet (previously it would `waitpid` for it to exit). Caveat emptor.                     

Since this bug only affects the parent, the child and grandchild share a memory space. Neither use pthreads and both exit quickly (and through different paths) so it should be fine.

Note that this will re-raise the slowdowns observed with large memory spaces. Caveat emptor.

mwkrentel:
Note: although it may be a little sloppy to ignore the return code for pthread_kill(),
it doesn't really matter.  In either case, monitor will be unable to signal the thread
to run code in the thread.  So, you'll still necessarily get some message that means
"can't connect to some threads" and you'll still miss the fini_thread callback.

I've never liked the need for signals for this case.  My plan for monitor NG, if it ever
happens, is not to call fini_thread for this case, instead just tell hpcrun that the process
is ending and let hpcrun deal with it.  For example, some threads will write out other
threads data, but not have this strong need to do everything from the same thread.

mwkrentel:
This (and related clone issues) would make great candidates for regression tests.

PR #470, opened 2021-10-27
Patch memory pool create

dejangrubisic:
This patch introduces **hpcrun_thread_init_mem_pool_once** at the beginning of each thread created from the monitoring tool (hpcrun or vendor-api) to process tracing/profiling data. This is necessary because these threads need to have initialized thread-data structure, since they create channels and other things with td data structure.

I added this function at the beginning of **cupti/rocm_buffer_completion_callback** and **gpu_trace_record**. I can see that there are pthread_create calls for:
- stream_data_collecting,
- finalize_all_thread_data,
- gpu_operation_record 

Should I put **hpcrun_thread_init_mem_pool_once** for them too?
Let me know if this patch solves the problem with cuda versions, we talked on the meeting.

### DIFF
--- a/src/tool/hpcrun/gpu/amd/roctracer-api.c
+++ b/src/tool/hpcrun/gpu/amd/roctracer-api.c
@@ -471,7 +471,7 @@ roctracer_buffer_completion_callback
  void* arg
 )
 {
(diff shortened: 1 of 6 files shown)


jmellorcrummey (approved):
LGTM

PR #475, opened 2021-11-01
Implement first-class placeholder support

### LINKED ISSUE #461: Design & implement first-class placeholder support
#### Background
There are a number of cases in call stacks where the actual functions involved are unimportant and so we only retain the semantic under a special name. Examples include the `<program root>` and `<thread root>`, `<gpu kernel>` and other GPU-related contexts, `<no activity>` and `<no thread>`. For the rest of this description these will be lumped together under the name "placeholders," which matches the name used in the Run codebase in many cases.

All placeholders are implemented as specially-named functions in `libhpcrun.so`, which are then translated into the appropriate `<...>` names during Prof. While this is simple, this also causes some headaches:
 - Measurements can't (easily) move between machines since the `libhpcrun.so` is no longer available to Prof.
 - Measurements must always be re-gathered for every version of Toolkit used if the previous one is deleted (by, say, `spack gc`). 
 - If Toolkit is reinstalled in-place (by, say, `make install`) the `libhpcrun.so` may not match the version gathered for the measurements (point 3 in https://github.com/HPCToolkit/hpctoolkit/issues/439).

In all of the above cases the symptom is that all placeholders are replaced with `<unknown procedure>` contexts, which is confusing and provides no information. The latter two cases are *really* easy to do when debugging out in the field.

#### Proposed Solution
Since placeholders don't have any meaning outside of their semantic (name), all that really needs to happen is for Run to use enumerated offsets in a fake load module (`0`) instead of `libhpcrun.so`. These can then be translated into something suitable in Prof2 and output the same way as before, just without the reliance on `libhpcrun.so`. This method is already used for `<partial call paths>` with good effect, it should be possible to extend this to the rest of the placeholders as well.

As an extension to the above thought, its really easy to add a touch of forward-compatibility by choosing the enumeration values to be 8-byte strings spelling a shorthand for the placeholder. If Prof2 or Proftt doesn't recognize a placeholder it can be given a name based on the shorthand, which will still be readable enough to a developer (as opposed to an opaque integer that needs to be looked up in a header file). (I got flak the last time I suggested this idea, my take is it's real easy to do and only provides benefits, so there's literally no point in not.)

#### Action Items
 - [ ] Add a shared `uint64_t` enumeration for all known placeholders in `prof-lean`, along with a function/table to look up their full names.
 - [ ] Convert all placeholder usage in Run to use the new placeholder enumerations. (Most should be easy, `<* root>` might be a bit trickier.)
 - [ ] Add a new `placeholder` Scope::Type with a single 8-byte value. For almost purposes acts like a `point`.
 - [ ] Update Hpcrun4 to translate offsets in module `0` as `placeholder` Scopes, passing the offset verbatim.
 - [ ] Update ExperimentXML4 to generate fake functions for `placeholder` Scopes, using their full names (or a fallback if unrecognized).

blue42u:
Resolves #461

There are many common scenarios where the libhpcrun.so or libmonitor.so used to perform measurement disappears or becomes out-of-date before Prof is run. This causes all "placeholder" contexts to show up as (at best) <unknown procedure> or (at worst) other random and unrelated functions. The affected placeholders include:
 - GPU operation placeholders (`<gpu *>`)
 - OpenMP operation placeholders (`<omp *>`)
 - `<no activity>`
 - CPU fences (`<program root>` and `<thread root>`)

This patch refactors and unifies all placeholders across Run into `prof-lean/placeholders.{h,c}`, which are now recorded as offsets in the special load module 0. This load module is never associated with a real binary, thus making placeholders unaffected by changes in the Toolkit installation.

For an additional touch of compatibility, the offsets used for placeholders are not contiguous and are instead comprised of 8-byte "shortcodes" (strings). These (vaugely readable) codes are used to generate fallback names for unrecognized values, allowing measurements to be passed between different versions of the placeholder set without losing easy interpretation by developers.

---

There are some changes to old-Prof to make it compile, they aren't and do not need to be tested or robust, Prof2 has the proper support and has been tested. There are also a handful of QoL features for placeholders, such as printing the shortcode and name as part of Proftt output.

Since there is a many-to-one mapping from IPs to placeholders, `hpcrun_denormalize_ip` cannot always give a proper answer (and in some cases, there simply is no answer). This function is only used in the OMPT interface for aesthetic minor offset adjustments, this has been altered to abort the adjustment if the IP was unable to be denormalized.


jmellorcrummey (approved):
LGTM

Issue #462, opened 2021-10-24
Suppressed threads have memory leaking issues


Jokeren:
https://github.com/HPCToolkit/hpctoolkit/blob/f1236e3521ca743f5d69bc878dbf0906df0a897f/src/tool/hpcrun/main.c#L778

When `demand_new_thread = true`, `hpcrun_threadMgr_data_get_safe` will allocate a new piece of data instead of using the existing ones in the memory pool. 

https://github.com/HPCToolkit/hpctoolkit/blob/f1236e3521ca743f5d69bc878dbf0906df0a897f/src/tool/hpcrun/main.c#L822

Then, in `hpcrun_thread_fini`, we don't call `hpcrun_threadMgr_data_put` to recycle the current thread's thread data. Thereby, a thread is both _suppressed_ and _demanding new thread data_ has the memory leak issue. This is benign usually. However, if an application creates thousands of threads, hpcrun will allocate a tens of gigabytes of CPU memory, causing an **serious problem**.

I observed this problem when monitoring laghos using the continuous pc sampling feature, which creates a helper thread at every `pcSamplingStart` call, and the thread is terminated at every `pcSampingStop` call. Since the helper threads are very lightweight, themselves do not cause performance issues. The real problem is that hpcrun allocates thread data but doesn't recycle it so that the memory usage is exploded from 1.9GB to 12GB within just 12 secs. If we ran Laghos for 10 minutes, it could consume ~600GB memory!

blue42u:
@Jokeren Is this not an observable issue now? I thought we still needed another patch making use of https://github.com/HPCToolkit/libmonitor/pull/2.

Jokeren:
@blue42u This patch in hpcrun could fix the problem, while the monitor change could make hpcrun even better by completely ignoring threads being monitored.

Jokeren:
@blue42u You are right. I also need libmonitor's support to get this problem fixed. Moreover, merging develop caused a bunch of other problems.

Jokeren:
#473 needs to be revised.

PR #473, opened 2021-10-29
Patch memory pool create

dejangrubisic:
I am making this PR as @Jokeren  asked me to get new changes to Develop, besides the master branch.

PR #477, opened 2021-11-03
Returning MONITOR_IGNORE_NEW_THREAD from monitor_thread_pre_create

dejangrubisic:
Patch memory pool create required one more change from Keren:
https://github.com/HPCToolkit/hpctoolkit/pull/473#issuecomment-958084480

Which is returning MONITOR_IGNORE_NEW_THREAD from monitor_thread_pre_create. This feature goes together with libmonitor ignore-from-pre-create branch that needs to be merged.
https://github.com/HPCToolkit/libmonitor/pull/2


blue42u (changes requested):
See below for the requested change.

Tested on Perlmutter (both alone and + requested change). No crash, thread ids did change but not significantly, no other glaring difference in measurements.

blue42u (approved):
All good now. Got word from Keren, this fixes https://github.com/HPCToolkit/hpctoolkit/issues/462.

Issue #432, opened 2021-09-21
Design & implement the taxonomy of metrics


blue42u:
#### Background
As of writing, `hpcrun -e gpu=nvidia,pc` measures over 80 metrics specific to CUDA code and contexts. All of these metrics are presented in the Viewer as separate metric columns, making browsing a database unwieldy and finding any particular metric more difficult than desired. Currently Prof2 makes this situation worse, by reassigning unique identifiers to metrics and thus removing the careful sorting of the metrics done in `hpcrun`. As such the metric columns in a Prof2-generated database are in an effectively random order. Although Prof2 can be fixed to retain the metric order, it is preferable to put time into a solution for the larger problem.

One such solution is to categorize the metrics in a structured fashion: users can browse with a few "high-level" metrics, which can be recursively expanded on request to show detailed "low-level" metrics when needed. In other words, the "high-level" metrics quickly (and concisely) tell you *where* the major issues are, while the "low-level" metrics give the details to tell *what* is the core symptom of the issue and (hopefully) *why* there's an issue in the first place.

#### Proposed Design
The Prof2 and Run sides are still under debate.

The proposed changes to the database format are as follows:
 - `<Metric>` tags within the `experiment.xml` will only contain `i` and `n` attributes, to be interpreted as previously. Metrics declared from `<Metric>` tags should **never** be presented to the user and instead serve to bind the canonical metric names to the ids used in the `thread.db` et al.
 - Databases will now contain a new `metrics/` subdirectory, reserved for metric taxonomy descriptions (plural in the event that Prof2 needs to generate more than one). Of these, only `metrics/default.yaml` will be "enabled" on database load, the remaining files in that directory may be "enabled" later by the user (in addition to other, user-provided files).
 - Each metric taxonomy file is a file written in standard YAML following a standard format that describes the metrics and their tree-like structure. A work-in-progress copy of the specification is [uploaded here](https://gist.github.com/blue42u/abf7d8a2750d9f8681218f7afd883b11). This format intends to address most of the issues required for practical metric taxonomy usage.

These "taxonomic metrics" have similar expressive power as derived metrics, however they additionally are laid in a tree arrangement, have additional rendering options, and do *not* specify visibility 

#### Action items
 - [X] Develop an initial format to transfer metric taxonomies to the Viewer.
 - [ ] Classify the currently available metrics into a taxonomy.
[...]

PR #481, opened 2021-11-11
[WIP] Metric Taxonomy

blue42u:
Initial, very WIP metric taxonomy support. Will eventually resolve https://github.com/HPCToolkit/hpctoolkit/issues/432.

Relative to the rough design proposed there, there are some additions to the [`<Metric>` tags in experiment.xml](https://github.com/HPCToolkit/hpctoolkit/blob/52818b64bae84f055a040a670e8b510e6a94c903/lib/dtd/hpc-experiment.dtd#L24-L45) and to the [input metrics in METRICS.yaml](https://github.com/HPCToolkit/hpctoolkit/blob/52818b64bae84f055a040a670e8b510e6a94c903/doc/METRICS.yaml#L21-L33). In short, measured metrics are now uniquely identified by a tuple of 4 strings:
 - `name`: arbitrary string used to name the exact metric that was measured,
 - `scope`: arbitrary string used to separate propagations (inclusive vs. exclusive, etc.),
 - `formula`: special-format string specifying the calculation applied to raw measurement values prior to combination, and
 - `combine`: string-enumeration specifying the operator to use for combinations.

In terms of the "accumulate" -> "combine" -> "finalize" view of analysis processing, `formula` specifies "accumulate" and `combine` specifies "combine," "finalize" comes from the formulas within the `METRICS.yaml`. All formulas are needed since non-summary data is raw and will need to be processed through these functions to generate a thread-local view with full metric taxonomy (i.e. miniature Prof on a small subset of the data). For this reason `<MetricDB>` tags are folded into the corresponding `<Metric>` tags, using a secondary identifier attribute `raw-i` to specify the id to use for non-summary profiles (to disambiguate the identifier attribute for the summary profile is now `summary-i`).

Currently this branch presumes Prof2 is in charge of "interpreting" the measured Metrics, rather than Run specifying what it wants to happen (in part because there is no clear answer to a "mixed" measurement with differing metrics in different threads). Whether this approach is sufficiently scalable remains to be seen.


src/lib/profile/sinks/metricsyaml.cpp
```diff
@@ -122,17 +118,469 @@ static YAML::Emitter& operator<<(YAML::Emitter& e, const MetricScope ms) {
...
+        for(const Metric& sm: sub_i) rawLeaf(out, sm);
+        out << EndSeq
+            << EndMap;
+      }
+      out << EndSeq
+          << EndMap;
+    }
+  } c_memcpy;
+
+  struct MemSet {
```
Jokeren, line 369: Do those structures need to be hard-coded? Or we can generate code parsing code from a configuration file (i.e., yaml or protocol buffer)?

blue42u: At the moment I haven't worked out how best to dynamically generate the taxonomic structure. I hard-coded the whole thing to get a minimal example together for experimental Viewer support. If you have an idea for an approach please elaborate,  suggestions would be very welcome.

aarontcopal2:
@blue42u , Is it possible to add an additional string for each metric called `format`? With that flexibility, one can represent metrics as floating-point or integer.

blue42u:
@aarontcopal2 `format` already exists in this PR? https://github.com/HPCToolkit/hpctoolkit/pull/481/files#diff-4598fc89d83c459cb8f8955220e821a27211179855ce63e059a89b4b282acd7eR114-R121

For integers I would recommend `%.3g`, that (should as per `printf`) render as integers until 1000 at which point it flips to sci-notation.

blue42u:
Rebased and superseded by #495.

PR #484, opened 2021-11-20
Add the E4S metadata

laksono:
The E4S metadata is part of E4S policy. My understanding is that it requires a yaml file in `.e4s` directory.
The content of the yaml is an adaptation from https://github.com/E4S-Project/E4S-Documenter/blob/master/data/hpctoolkit/e4s.yaml where the `MemberProduct` field is set to `True` (instead of `False`).

The metadata is not well documented and strangely, even E4S team doesn't know what this field means.


jmellorcrummey (approved):
LGTM

PR #486, opened 2021-11-29
ensure roctracer traces are written to disk

jmellorcrummey:
use a device flush finalizer to  ensure that roctracer activity
tracing is shut down and roctracer activity traces are
flushed.

this must be done before a device shutdown finalizer calls
gpu_trace_fini. otherwise, some trace records may not be flushed
to disk.

the handshake between gpu_trace_fini and tracing threads REQUIRES
that tracing has been shut down and traces are flushed from
roctracer before gpu_trace_fini sets stop_trace_flag.


mxz297 (approved):
LGTM

Issue #440, opened 2021-09-21
Refine the GPU CCT reconstruction bits.


Jokeren:
While I think they work just fine, I don't actually have the proper test cases or knowledge to know whether they are complete. I could use some confirmation they are sufficient and complete for all the crazy reconstructions we do.

They probably also need some code cleanup.

#### Action items
 - [ ] Properly document the non-standard Context classes and `Metric::*finalize`, so someone other than me (@blue42u) can actually understand what's going on.
 - [ ] (If needed) refactor CollaborativeContexts to use clean C++ instead of whatever mess it is currently.

Jokeren:
@blue42u check out test cases prefixed by `cu_call_path`. 
https://github.com/Jokeren/hpctoolkit-gpu-samples

blue42u:
@Jokeren I know those exist (and remember experimenting with a few when I wrote the pieces originally), but I have no idea what they're supposed to look like in the final database. I could compare with the current output from `master` but I don't understand the algorithm well enough to know what an acceptable error bound is, nor what I'm looking for to test the edge cases (e.g. all-zeros distribution). And its unclear to me whether any of these can test the infrastructure for range-based sampling, which I'm less confident in.

So, if you could, I would appreciate it if you test the current Prof2 and let me know how it fares. That request extends to the Intel and AMD support projects as well (@aarontcopal2, @dejangrubisic and @mxz297 I think?), I really just need more information on what Prof2 needs to provide to support everyone's GPU work for this front.

Jokeren:
Those tests do not cover range-based sampling because they only have a single GPU kernel. I'll try the current prof2 first and see if it works

Jokeren:
Actually, if hpcstruct could mark the GPU structure file as `without cfg`, I feel like hpcprof2's check could be simplified.

blue42u:
(For context, me and Keren debugged his issue and found it was a pilot error, he forgot the `--gpucfg yes`.)

@Jokeren You've got the right idea, the marker needs to be a tristate though (`--gpucfg no`, `--gpucfg yes` and not a GPU binary or a GPU binary without Dyninst support). The other factor is that if you've measured with `-e gpu=nvidia` you can't use CFG data anyway so we shouldn't suggest it. I currently can't think of an (agnostic) way to detect that easily, it would help if the CCT used to mark the kernel function in Run was marked as a "function-level CCT" (as opposed to the usual instruction-level CCTs).

PR #493, opened 2021-12-08
Calling Context Reconstruction Redesign

blue42u:
Redesigns how GPU calling context reconstruction works (in Prof2) from the ground up. Resolves https://github.com/HPCToolkit/hpctoolkit/issues/440.

Apparently no one (including me sometimes) fully understood how GPU calling context reconstruction worked. It's not an easy problem, but this PR rewrites it from the ground up with (hopefully) a better conceptual framework, and then cleans up the surrounding parts of the Prof2 core. There are also longer comments to help understanding the purposes of the new classes. See the commit log for details on exactly how things changed.

In addition to redesigning Context-likes, this also removes the ProfileTransformers and Classification, folding their roles into the ProfileFinalizers (+ their custom userdata). As our analysis is evolving it is making less and less sense to force all binary analysis into a single common data structure, having each Finalizer decide is much cleaner. Some new template classes in `hpctoolkit::util` are added to assist with that.

This PR slightly changes the `.hpcrun` format to add a 1-byte bitfield to the CCT nodes, a flag in that bitfield indicates whether a CCT node needs reconstruction or not. This allows Prof2 to know immediately what variety of Context will be created by a CCT node, without referring to Structfiles or other external information. This change has no backwards-compatibility, all measurements will need to be redone.

@Jokeren: I would appreciate it if you could test this branch with your test cases, I'm more confident in the formulas but not my testing. The `new-cupti-context-rewrite` branch is (an old) `new-cupti` plus a port of the new `.hpcrun` format change, you can use that (possibly rebased) to generate some measurements.


Jokeren:
@blue42u I'll try this branch soon. Thanks for your effort.

Can you please tell me which functions I need to modify to apply new the apportion mechanism?

blue42u:
If you mean the `GKER:SAMPLED_COUNT` support, it should already be implemented in this branch: https://github.com/HPCToolkit/hpctoolkit/blob/pr-prof2-context-rewrite/src/lib/profile/finalizers/struct.cpp#L243-L246

The apportion mechanism (or rather, the generation of factors) is implemented in [context.cpp](https://github.com/HPCToolkit/hpctoolkit/blob/pr-prof2-context-rewrite/src/lib/profile/context.cpp), it takes up most of the file. The apportioning gets applied in the first part of [`PerThreadTemporary::finalize`](https://github.com/HPCToolkit/hpctoolkit/blob/d4c29bdce16532f726da812a31323f8163b72243/src/lib/profile/accumulators.cpp#L105) using those factors.

Jokeren:
@blue42u Thanks for pointing it out. Let me check if it works, looks like `total_val_exteriorLogical` and `total_val_exterior` handle something I want.

blue42u:
Changed:
 - Fixed a number of bugs @Jokeren found, including removing sanity checks that weren't necessary and weren't correct.
 - Renamed "from ununwindable" to just (not) "unwound" for the CCT-related bits.

Issue #460, opened 2021-10-23
Fix the ExperimentXML4 CCT output


blue42u:
#### Background
Prof2 outputs an experiment.xml (version 4, EXMLv4 for short) file in every database, intended to be similar in structure to the (version 2) file output by old-Prof. These contain information about the database including the unified calling context tree, the format for this section is (mostly) unchanged from version 2.

Unfortunately the internal Prof2 representation for context ([loosely documented here](https://github.com/HPCToolkit/hpctoolkit/blob/develop/src/lib/profile/scope.hpp#L94-L106)) is very different from the representation used in EXMLv2. The latter is basically undocumented (AFAIK, and I've been asking for a while now) so most of the logic in ExperimentXML4 is based on my own experimentation trying to get the Viewer to behave. Which is a moving target since @laksono kindly has been tweaking the Viewer to work around Prof2's quirks. This does not seem like a stable or long-lived approach.

The Prof2 representation and the Viewer presentation have some similarities, most Prof2 Contexts correspond to a single context row in the Viewer. The main differences are the smallest granularity (instruction in Prof2, line in the Viewer) and the handling for `<unknown procedure>` (Viewer present an extra "dummy" context row, Prof2 just closely nests `point` Contexts). In short, the organically designed EXML format in between Prof2 and the Viewer is quite possibly a significant detour.

#### Proposed Solutions
One solution is to solidify the EXMLv4 CCT format (as the Viewer understands it) and change Prof2 to satisfy that. This direction requires proper documentation for the entire format in the FORMATS.md ([here](https://github.com/HPCToolkit/hpctoolkit/blob/7ac59f69a6ee83b2868e222586266cdb69737e4a/doc/FORMATS.md?plain=1#L101)), including tag and attribute meanings and cross-tag interactions. This also requires documentation on how this format is presented in the Viewer (including the weird "special cases" added to work around old-Prof bugs), once documented this handling should be relatively static.

This places the responsibility of converting code semantics to intended presentations on Prof2, which would "squish" its internal representation into the EXMLv4 format to generate the intended presentation. The benefit of this approach is that new types of Scope can be added with only changes in Prof2. The downside is that the EXML **will** be a strange output with very little semantic meaning outside of its presentation in the Viewer, which might complicate external readers (Hachet) or new Viewer-only analyses (next gen after GPU blame shifting).

**--OR--**

[...]

PR #495, opened 2021-12-20
Refactor database "metadata": meta.db

blue42u:
Refactor the "metadata" format for databases, previously known as `experiment.xml` and now affectionately called `meta.db`. Supersedes https://github.com/HPCToolkit/hpctoolkit/pull/481. Resolves ~~https://github.com/HPCToolkit/hpctoolkit/issues/432~~ and https://github.com/HPCToolkit/hpctoolkit/issues/460. Depends on https://github.com/HPCToolkit/hpctoolkit/pull/514.

There are a number of changes that we want to make to the database "metadata," stored in experiment.xml:
 - Metrics are growing more complex and numerous, we need a new taxonomic structure above the raw measured + propagated data to better present metric data (https://github.com/HPCToolkit/hpctoolkit/issues/432).
 - The representation for calling context is rough and difficult for Prof2 to represent clearly, we need a new format for the CCT that better matches how the analysis is actually performed and presented (https://github.com/HPCToolkit/hpctoolkit/issues/460).
 - XML is slow to parse and our shortening of names and identifiers makes it only vaguely readable, we need a compact binary representation for database "metadata."

Since these all are breaking changes and require new versions of the Viewer, it makes much more sense to do all the changes "at once" with one PR rather than over a series of PRs. This PR and branch is the "one PR."

The binary format to replace experiment.xml is `meta.db`, and is designed to retain most of the same structure and functionality as experiment.xml with the adjustments required by the other two changes. It is also extensible and is designed to support both backwards compatibility and limited forward compatibility (some things won't look right, but nothing will be completely missing and the file *will* parse to completion).

Currently this branch presumes Prof2 is in charge of "interpreting" the measured Metrics, rather than Run specifying what it wants to happen (in part because there is no clear answer to a "mixed" measurement with differing metrics in different threads). Whether this approach is sufficiently scalable and robust remains to be seen.

While this is a work-in-progress, the changes to the formats are documented in the expected place: [doc/FORMATS.md](https://github.com/HPCToolkit/hpctoolkit/blob/pr-metadb-refactor/doc/FORMATS.md) and the new [doc/METRICS.yaml](https://github.com/HPCToolkit/hpctoolkit/blob/pr-metadb-refactor/doc/METRICS.yaml).


blue42u:
[This repository has moved to GitLab, please continue this PR there.](https://gitlab.com/hpctoolkit/hpctoolkit/-/issues/495)

blue42u:
The previous link was incorrect, apologies. The corrected link is [here.](https://gitlab.com/hpctoolkit/hpctoolkit/-/merge_requests/495)

PR #501, opened 2022-01-12
Update the man page of the viewer and its download page

laksono:
Minor updates:
- Fix and refine the man page for the viewer
- Update the download page to show the prior versions (2021.10 and 2021.05)


martyitz:
I don't understand the deletion of the "Show color mapping" reference.  Is that control now gone?
Otherwise, it looks OK to me.

martyitz:
Done.

	Marty

laksono:
> I don't understand the deletion of the "Show color mapping" reference. Is that control now gone? Otherwise, it looks OK to me.

It's my mistake. It should be just "Color mapping" (no "Show"). I am not sure why it's now deleted. I'll fix it.
Thanks for the correction!

PR #503, opened 2022-01-13
fix cupti thread and process finalizer order

jmellorcrummey:
- ensure that device finalizers are called in the order that they are registered
- register cupti_device_flush as a device_flush finalizer
- regster cupti_device_shutdown and then gpu_trace_fini as device_shutdown finalizers


Jokeren:
Looks good to me

Jokeren (approved):

PR #504, opened 2022-01-14
roctracer support for fixing finalization

jmellorcrummey:
- thread device finalizer flushes trace
- process device finalizer shuts down rocm measurement, flushes trace, then calls gpu_trace_fini

### DIFF
--- a/src/tool/hpcrun/gpu/amd/roctracer-api.c
+++ b/src/tool/hpcrun/gpu/amd/roctracer-api.c
@@ -578,6 +578,17 @@ roctracer_init
 }
 
 void
+roctracer_flush
+(
+ void* args,
+ int how
+)
+{
+  HPCRUN_ROCTRACER_CALL(roctracer_flush_activity_expl, (NULL));
+}
+
+
+void
 roctracer_fini
 (
  void* args,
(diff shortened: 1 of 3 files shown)


jmellorcrummey:
my fix at the end of november got the order between flush and gpu_trace_fini right. however, every thread executed the complete rocm finalizer, which shuts down rocm and flushes. 

with this change, each thread flushes.

main thread flushes, shuts down rocm, flushes again (just to be sure), and then calls gpu_trace_fini.

src/tool/hpcrun/gpu/amd/roctracer-api.c
```diff
@@ -577,6 +577,17 @@ roctracer_init
...
 
+void
+roctracer_flush
+(
+ void* args,
+ int how
+)
+{
+  HPCRUN_ROCTRACER_CALL(roctracer_flush_activity_expl, (NULL));
+}
```
mxz297, line 588: @jmellorcrummey I have the feeling that we should also move the call to `gpu_application_thread_process_activities()` in `roctracer_fini` to `roctracer_flush`. Otherwise, child threads will flush roctracer activities without process them?

jmellorcrummey: agreed. done.

PR #505, opened 2022-01-14
finalize level0 traces

jmellorcrummey:
finalize level0 traces by calling gpu_trace_fini as last device finalizer when shutting down the process.


mxz297 (approved):
I don't think Level0 tracing currently works, but I agree that this change should be there and is necessary for Level0 tracing to work eventually.

jmellorcrummey:
> I don't think Level0 tracing currently works, but I agree that this change should be there and is necessary for Level0 tracing to work eventually.

That was my thought exactly.  I fixed it while I was thinking about it.

PR #509, opened 2022-01-21
CUDA-11.6-fix

Jokeren:
Fix a cuda cfg parsing problem for CUDA >= 11.5: target labels are changed from .L_<number> to .L_x_<number>, so new code is added to extract name using the new pattern.

Interestingly, using cuda-11.6 collects 13% less samples than cuda-11.4 for quicksilver, but it has nothing to do with postmortem analysis. I confirmed this issue using hpctoolkit's log file.


mxz297 (approved):
LGTM

Jokeren:
@jmellorcrummey Thanks for the suggestion. Please check the simplified code.

PR #512, opened 2022-01-24
OMPT support for AOMP on AMD GPUs

jmellorcrummey:


jmellorcrummey:
the ompt-amd branch was already merged into develop and master

PR #513, opened 2022-01-30
Quality of Life Patches

blue42u:
Handful of patches to fix and update questionable code, some of which are required fixes for active development.

This currently includes:
 - Replacements for the CCT iteration algorithms that do not use stack recursion. This version of the patch uses a single pointer, fixes some bugs, and is simpler and better documented. This fixes difficulties @Jokeren has reported working with CCT nodes with thousands of children.
 - Adjustments to `hpcrun_safe_enter/exit` usage to support nested "safety," which now includes the entirety of the CUPTI subscriber callback. This fixes difficulties I saw previously where unwinding through the callback + libcuda.so would fail (as it often does) and cause the whole program to hang (possibly related to using the `*MSG` macros outside of the "safe" regions).
 - A new `-ck MAX_UNWIND_DEPTH` argument for controlling the maximum number of unwind-step attempts an unwind will make before giving up and aborting the sample. Defaults to 1000. Does not fix any actively seen bugs.
 - Replacements for some uses of `assert` explicit `abort`s, since the former doesn't stop the program when `NDEBUG` is defined. The ones replaced are ones I could not determine were independent of external factors. This is required for Meson support, but we could add it to the Autogoo with very little effort and possibly a few optimization gains.
 - Fixes for the Valgrind annotations support, in part by copying the client headers provided by Valgrind into our codebase. These headers are under a weaker (BSD-style) license to support this very thing. This is required for using the Valgrind annotations on systems where the headers aren't in the right spot (I had that happen at least once).
 - **ADD:** Tweak to `HPCRUN_ABORT_TIMEOUT` to transfer control to a registered `SIGABRT` handler before dying. This helps debugging on systems where coredumps aren't available but the MPI provides a backtrace on various signals.


src/tool/hpcrun/os/catamount/override.c
```diff
@@ -47,6 +47,7 @@
 #include <assert.h>
```
jmellorcrummey, line 47: we can delete this file and the directory containing it. catamount was a predecessor to cray's compute node linux.

src/tool/hpcrun/sample-sources/papi-c.c
```diff
@@ -154,7 +154,7 @@ get_event_index(sample_source_t *self, int event_code)
     int evcode = self->evl.events[i].event;
     if (event_code == evcode) return i;
   }
-  assert(0);
+  assert(false && "Invalid event_code!");
```
jmellorcrummey, line 157: no abort?

src/tool/hpcrun/cct/cct.c
```diff
@@ -222,28 +225,62 @@ splay(cct_node_t* cct, cct_addr_t* addr)
...
+// See splay_walk_next for more details.
+static cct_node_t* splay_walk_init(cct_node_t* root) {
+  if(root == NULL) return NULL;
+  assert(root->up == NULL);
 
-  walk_child_lrs(cct->left, op, arg, level, wf);
-  walk_child_lrs(cct->right, op, arg, level, wf);
-  wf(cct, op, arg, level);
+  // First node in the iteration is always the leftmost child.
+  return splay_leftmost_descent(root);
```
jmellorcrummey, line 252: why splaying at each level of the walk?

I'm going to need to discuss the rationale for this with @blue42u

src/tool/hpcrun/unwind/common/backtrace.c
```diff
@@ -170,6 +171,8 @@ hpcrun_skip_chords(frame_t* bt_outer, frame_t* bt_inner,
   return &bt_inner[skip];
 }
 
+static _Thread_local int max_unwind_attempts = -1;
```
jmellorcrummey, line 174: Why is this thread local rather than just a local variable in the backtrace procedure?

src/tool/hpcrun/unwind/common/backtrace.c
```diff
@@ -194,6 +197,9 @@ hpcrun_generate_backtrace_no_trampoline(backtrace_info_t* bt,
 
   step_state ret = STEP_ERROR; // default return value from stepper
 
+  if(max_unwind_attempts < 0)
```
jmellorcrummey, line 200: shouldn't we set this once in an init procedure outside the backtrace routine?

src/tool/hpcrun/unwind/common/backtrace.c
```diff
@@ -266,6 +274,13 @@ hpcrun_generate_backtrace_no_trampoline(backtrace_info_t* bt,
       ret = hpcrun_unw_step ( &cursor, &steps_taken);
     }
 
+    // If we've "unwound" too many frames, probably something is wrong.
+    attempts++;
+    if(attempts > max_unwind_attempts) {
+      EMSG("Unwind took too many attempts, aborting unwind (taken: %d > %d)", steps_taken, last_steps_taken);
```
jmellorcrummey, line 280: I don't understand the roles of steps_taken and last_steps_taken.

jmellorcrummey (commented):
I will need to discuss the CCT iterator before approving.

Is abort the right thing to call instead of monitor_real_abort? (Unless we overrode assert, that was the behavior before.)

I have some questions about limiting the backtrace.

src/tool/hpcrun/main.c
```diff
@@ -413,6 +401,7 @@ abort_timeout_handler(int sig, siginfo_t* siginfo, void* context)
   long pid = (long) getpid();
   EEMSG("hpcrun: abort timeout activated in process %ld - context pc %p", 
     pid, hpcrun_context_pc(context)); 
+  abort();
```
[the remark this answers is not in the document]
blue42u: `monitor_real_abort()` will skip over whatever signal handlers are registered for `SIGABRT`. On systems where coredumps aren't easily accessible (Summit, Perlmutter, ...) doing this lets the MPI produce a stack dump, which is at least better than nothing.

(If I'm just blind and there's a way to get coredumps on these systems correct me, but AFAICT the cores get ferried away behind sysadmin permissions.)

In my experience using just `abort()` still generates a coredump in precisely the right spot when coredumps are available, libmonitor intercepts but then immediately re-raises the `SIBABRT` for the desired effect.

src/tool/hpcrun/main.c
```diff
@@ -1264,14 +1253,13 @@ monitor_init_thread(int tid, void* data)
...
 
-  hpcrun_safe_enter();
-
   TMSG(THREAD,"init thread %d",tid);
   void* thread_data = hpcrun_thread_init(tid, (local_thread_data_t*) data, ! hpcrun_thread_suppress_sample);
   TMSG(THREAD,"back from init thread %d",tid);
 
   hpcrun_threadmgr_thread_new();
 
+  // The new thread is initialized as "inside hpcrun", so we have to switch back
```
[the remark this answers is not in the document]
blue42u: The issue is that `hpcrun_safe_enter()` returns false if the `thread_data_t` isn't available yet, so any safety nests above this function will always be lost. This (unconditional) call is needed to make sure we are "outside hpcrun" before the application fires up and we start getting real data.

The alternative is to make the `inside_hpcrun` flag a thread-local variable instead of binding it to the `td`, that way it would always work (presuming we don't jump in before our static TLS initialization... which I'm not yet confident is impossible).

src/tool/hpcrun/cct/cct.c
blue42u: `splay` as in "splay tree," this function doesn't alter the tree's shape, it only fiddles with the `up` field.

Specifically, `spack_leftmost_descent` returns the leftmost lowest child of the splay tree rooted at `root`, setting the `up` fields to allow for travel back up the tree when needed.

blue42u:
[This repository has moved to GitLab, please continue this PR there.](https://gitlab.com/hpctoolkit/hpctoolkit/-/issues/513)

blue42u:
The previous link was incorrect, apologies. The corrected link is [here.](https://gitlab.com/hpctoolkit/hpctoolkit/-/merge_requests/513)

PR #518, opened 2022-02-10
New cupti

Jokeren:
@blue42u As this branch is growing huge, I'd like to merge it into develop before adding new stuff.

I'll do some code clean up first. My plan is to use the `--enable-new-cupti` flag to activate new code. If `--enable-new-cupti` is not use, the same CUPTI code as develop will be executed.


Jokeren:
Also, @blue42u Can you take a look at the conflict in metrc.cpp. Is it because some commits have not been merged into develop yet? https://github.com/HPCToolkit/hpctoolkit/pull/518/conflicts

blue42u:
@Jokeren This branch is *faaaar* too much of a mess to merge, the only reason you have conflicts is because you have leftover Prof2 history that has nothing to do with the new CUPTI support in this branch. 

So, I would recommend you squash out all the history, prune the diff down to just the parts that you need for the new CUPTI interface, and rebase that onto the latest `develop`. Then we can start merging.

PR #519, opened 2022-02-10
[New CUPTI] Main

Jokeren:
### Test

`-ck HPCRUN_CUDA_NEW_CUPTI=TRUE` should be applied to enable fast unwinding and range profiling.

#### Regression

*PeleC/PMF* stop_time=100, max_time=200, 6.3s

- [x] `hpcrun -e gpu=nvidia` 9.0s
- [x] `hpcrun -e gpu=nvidia,pc` 104s

#### New CUPTI

*PeleC*

The following commands should generate databases with the same quality as if without `HPCRUN_CUDA_NEW_CUPTI=TRUE`, except that coarse-grained metrics (e.g., `GXCOPY`) are not available in the fine-grained profiling mode (i.e., `gpu=nvidia,pc`).

- [x] `hpcrun -e gpu=nvidia -ck HPCRUN_CUDA_NEW_CUPTI=TRUE` 8.5s
- [x] `hpcrun -e gpu=nvidia,pc -ck HPCRUN_CUDA_NEW_CUPTI=TRUE` 86s

The following commands use range profiling or fast unwinding to reduce overhead.

- [x] `hpcrun -e gpu=nvidia -ck HPCRUN_CUDA_NEW_CUPTI=TRUE -ck HPCRUN_CUDA_FAST_UNWIND=TRUE` 7.5s
- [x] `hpcrun -e gpu=nvidia,pc -ck HPCRUN_CUDA_NEW_CUPTI=TRUE -ck HPCRUN_CUDA_RANGE_MODE=CONTEXT_SENSITIVE` 44s
- [x] `hpcrun -e gpu=nvidia,pc -ck HPCRUN_CUDA_NEW_CUPTI=TRUE -ck HPCRUN_CUDA_RANGE_MODE=TRIE` 44s

For unknown reasons, `-ck CUDA_SYNC_YIELD=TRUE` does not have any effect on PeleC. We had to modify PeleC source code to change the synchronization policy.

#### Limitations

* Coarse-grained metrics and fine-grained metrics are profiled separately. Since kernel metrics such as the number of register usage are no longer available in the fine-grained profiling mode, we cannot derive SM efficiency and occupancy metrics. 
* Profiling an application without any GPU kernel will hang using the range profiling model.
* [CUPTI Bug] CUPTI PC Sampling API slows down cudaDeviceSynchronize().
* [CUPTI Bug] CUPTI PC Sampling records overflow after profiling a range with many big kernels.
* [CUPTI Bug] CUPTI incurs high profiling overhead (i.e., 1.5x) for Laghos.
* [CUPTI Limitation] CUPTI PC Sampling API does not support profiling multiple contexts using the continuous (i.e., range) mode.


src/lib/banal/Struct.cpp
```diff
@@ -150,7 +151,7 @@ using namespace std;
 #define DEBUG_SKEL_SUMMARY  0
 
 #if DEBUG_CFG_SOURCE || DEBUG_MAKE_SKEL || DEBUG_SHOW_GAPS
-#define DEBUG_ANY_ON  1
+#define DEBUG_ANY_ON  0
```
blue42u: Other side of the conditional is identical. Do you mean to delete the `DEBUG_ANY_ON` usage?

src/lib/binutils/intel/gen_binary_decoder.h
```diff
@@ -29,7 +29,11 @@
 #include <string>
 
 
+<<<<<<< HEAD
+#ifdef OPT_ENABLE_IGC
+=======
 #ifdef ENABLE_IGC
+>>>>>>> 770ebcaf0a2f06f5dac49b98bc9e822e9c40797a
```
blue42u: Leftover conflict in unrelated code region, remove.

src/lib/binutils/intel/gen_binary_decoder.h
```diff
@@ -74,6 +78,10 @@ class GenBinaryDecoder {
  private:
   KernelView kernel_view_;
 };
+<<<<<<< HEAD
+#endif // OPT_ENABLE_IGC
+=======
 #endif // ENABLE_IGC
+>>>>>>> 770ebcaf0a2f06f5dac49b98bc9e822e9c40797a
```
blue42u: Leftover conflict in unrelated code region, remove.

src/lib/prof-lean/crypto-hash.h
```diff
@@ -159,6 +159,10 @@ crypto_hash_self_test
   int verbose
 );
 
+#if defined(__cplusplus)
+}
+#endif
+
```
blue42u: Unrelated code region, would break the build if self-test was switched on. Remove.

src/lib/prof-lean/hpcrun-fmt.c
```diff
@@ -820,10 +820,10 @@ hpcrun_fmt_lip_fprint(lush_lip_t* x, FILE* fs, const char* pre)
   return HPCFMT_OK;
 }
 
-
```
blue42u, line 823: Unrelated whitespace changes (more in file). Remove all.

src/lib/profile/scope.cpp
```diff
@@ -57,6 +57,7 @@
 #include <iomanip>
 #include <stdexcept>
 #include <sstream>
+#include <iomanip>
```
blue42u, line 60: Unrelated, remove. If there's a compile failure in this file because of it let me know.

configure.ac
```diff
@@ -669,7 +669,6 @@ else
   my_opt_cxxflags="-g -O3 -Wall"
 fi
 
-
```
blue42u, line 672: Loose whitespace change. Remove.

configure.ac
```diff
@@ -815,6 +814,7 @@ case "${host}" in
     ;;
 esac
 
+
```
blue42u, line 817: Loose whitespace change. Remove.

src/lib/prof-lean/crypto-hash.h
```diff
@@ -51,7 +51,7 @@
 //
 // Purpose:
 //   compute a cryptographic hash of a sequence of bytes. this is used
-//   to name information presented to hpcrun in memory (e.g. a GPU binary) 
+//   to name information presented to hpcrun in memory (e.g. a GPU binary)
```
blue42u, line 54: (Whole file): No consequential changes. Remove.

src/lib/binutils/intel/gen_binary_decoder.h
```diff
@@ -7,10 +7,10 @@
 // to use, copy, modify, merge, publish, distribute, sublicense, and/or sell
 // copies of the Software, and to permit persons to whom the Software is
 // furnished to do so, subject to the following conditions:
-//
```
blue42u, line 10: (Whole file): No consequential changes. Remove.

src/lib/banal/Struct.cpp
```diff
@@ -63,42 +63,21 @@
 
 //***************************************************************************
 
-#include <sys/types.h>
```
blue42u, line 66: (Whole file): Not certain if there are consequential changes. Please minimize.

src/tool/hpcprof2-mpi/mpi.cpp
```diff
@@ -103,7 +103,6 @@ const Op& Op::min() noexcept { return o_min; }
 static BaseOp o_sum = MPI_SUM;
 const Op& Op::sum() noexcept { return o_sum; }
 
-
```
blue42u, line 106: Unrelated, remove.

src/tool/hpcrun/audit/auditor.c
```diff
@@ -341,11 +346,51 @@ uintptr_t la_symbind32(Elf32_Sym *sym, unsigned int ndx,
...
+  //  exports.cuMemcpyDtoHAsync = sym->st_value;
+  //  return hooks.cuMemcpyDtoHAsync;
+  //} else if (strcmp(symname, "cuMemcpyHtoD") == 0) {
+  //  exports.cuMemcpyHtoD = sym->st_value;
+  //  return hooks.cuMemcpyHtoD;
+  //} else if (strcmp(symname, "cuMemcpyHtoDAsync") == 0) {
+  //  exports.cuMemcpyHtoDAsync = sym->st_value;
+  //  return hooks.cuMemcpyHtoDAsync;
+  //}
+
```
blue42u: Code in comments. Remove.

src/tool/hpcrun/audit/audit-api.h
```diff
@@ -99,6 +100,63 @@ typedef struct auditor_hooks_t {
...
+  CUresult (*cuMemcpyHtoD)(
+    CUdeviceptr dstDevice,
+    const void* srcHost,
+    size_t ByteCount);
+
+  CUresult (*cuMemcpyHtoDAsync)(
+    CUdeviceptr dstDevice,
+    const void* srcHost,
+    size_t ByteCount,
+    CUstream hStream);
```
blue42u: This and related changes in `hpcrun/audit/` appear to be dead. Remove?

src/tool/hpcrun/cct/cct.c
```diff
@@ -68,36 +68,32 @@
 //************************* System Include Files ****************************
 
 #include <assert.h>
+#include <stdbool.h>
```
blue42u, line 71: (Whole file): Lots of inconsequential changes. Please minimize.

src/tool/hpcrun/cct/cct.c
```diff
@@ -68,36 +68,32 @@
...
+#include "../memory/hpcrun-malloc.h"
 #include "cct.h"
-#include "cct_addr.h"
 #include "cct2metrics.h"
-#include "../memory/hpcrun-malloc.h"
+#include "cct_addr.h"
 //#include "../ompt/ompt-interface.h"
 //#include "../memory/hpcrun-malloc.h"
 
-#define HPCRUN_CCT_KEEP_DUMMY 1
```
blue42u, line 95: This and related: seemingly unrelated change. Remove or factor into a separate commit.

src/tool/hpcrun/cct/cct.h
```diff
@@ -102,7 +102,6 @@
 #define ADDR(L)      (cct_addr_t) NON_LUSH_ADDR_INI(L)
 #define ADDR2_I(id, ip) NON_LUSH_ADDR_INI(id, ip)
 #define ADDR2(id, ip) (cct_addr_t) ADDR2_I(id, ip)
-#define HPCRUN_DUMMY_NODE 65534
```
blue42u, line 105: Seemingly unrelated change, remove or factor out.

(Rest of file): Inconsequential changes. Remove.

src/tool/hpcrun/control-knob.c
```diff
@@ -38,12 +38,16 @@ control_knob_register(char *name, char *value, control_knob_type type)
...
     control_knobs = iter;
   }
-  iter->value = strdup(value);
-  
+  if (iter->type == ck_string) {
+    iter->value = strdup(value);
+  } else {
+    iter->value = (char *) malloc(sizeof(char) * (strlen(value) + 1));
+    strcpy(iter->value, value);
+  }
```
blue42u, line 50: Completely inconsequential change. Remove.

src/tool/hpcrun/control-knob.c
```diff
@@ -62,20 +66,18 @@ control_knob_init()
   char *in = getenv("HPCRUN_CONTROL_KNOBS");
   if (in == NULL) return;
 
-  char *save;
+  char *save = NULL;
   for (char *f = start_tok(in); more_tok(); f = next_tok()){
     char *tmp = strdup(f);
     char *name = strtok_r(tmp, "=", &save);
-    control_knob_t *iter = NULL;
+    char *value = strtok_r(NULL, "=", &save);
```
blue42u, line 73: This does not look right at all.

src/tool/hpcrun/control-knob.h
```diff
@@ -1,7 +1,7 @@
 #ifndef HPCRUN_CONTROL_KNOB_H
 #define HPCRUN_CONTROL_KNOB_H
 
-typedef enum{
+typedef enum {
```
blue42u, line 4: Unrelated whitespace change. Remove.

src/tool/hpcrun/gpu/gpu-activity-process.c
```diff
@@ -47,26 +47,24 @@
 
 #include <assert.h>
 
-
-
```
blue42u, line 51: (Whole file): Not sure if there are consequential changes. Please minimize.

src/tool/hpcrun/gpu/instrumentation/gtpin-instrumentation.c
```diff
@@ -121,8 +121,6 @@ static __thread uint64_t gtpin_cpu_submit_time = 0;
 static __thread gpu_op_ccts_t gtpin_gpu_op_ccts;
 static __thread bool gtpin_first = true;
 
-
-
```
blue42u, line 125: (Whole file): Unrelated inconsequential changes. Remove.

src/tool/hpcrun/main.c
```diff
@@ -845,16 +847,17 @@ hpcrun_thread_fini(epoch_t *epoch)
...
     hpcrun_id_tuple_cputhread(td);
 
+    // add separator for each compact thread
+    bool add_separator = true;
     hpcrun_threadMgr_data_put(epoch, td, add_separator);
-
-    TMSG(PROCESS, "End of thread");
   }
+
+  TMSG(PROCESS, "End of thread");
```
blue42u: Unrelated change? Remove or rationalize.

src/tool/hpcrun/main.c
```diff
@@ -845,16 +847,17 @@ hpcrun_thread_fini(epoch_t *epoch)
...
     // thread manager may enqueue the thread_data (in compact mode)
     // or flush the data into hpcrun file
-    int add_separator = 0;
     thread_data_t* td = hpcrun_get_thread_data();
 
     // assign id tuple for pthreads
     hpcrun_id_tuple_cputhread(td);
 
+    // add separator for each compact thread
+    bool add_separator = true;
```
blue42u: Unrelated change? Remove or rationalize.

This is generic code, what consequence does this have on pure-CPU profiles?

src/tool/hpcrun/main.c
```diff
@@ -1746,13 +1762,125 @@ static void auditor_init() {
...
   hooks->stable = auditor_stable;
+  //hooks->cuLaunchKernel = auditor_cuLaunchKernel;
+  //hooks->cuMemcpy = auditor_cuMemcpy;
+  //hooks->cuMemcpyAsync = auditor_cuMemcpyAsync;
+  //hooks->cuMemcpyDtoH = auditor_cuMemcpyDtoH;
+  //hooks->cuMemcpyDtoHAsync = auditor_cuMemcpyDtoHAsync;
+  //hooks->cuMemcpyDtoD = auditor_cuMemcpyDtoD;
+  //hooks->cuMemcpyDtoDAsync = auditor_cuMemcpyDtoDAsync;
+  //hooks->cuMemcpyHtoD = auditor_cuMemcpyHtoD;
+  //hooks->cuMemcpyHtoDAsync = auditor_cuMemcpyHtoDAsync;
```
blue42u: Code in comments, remove before merging.

src/tool/hpcrun/main.c
```diff
@@ -1746,13 +1762,125 @@ static void auditor_init() {
...
+  CUdeviceptr dstDevice,
+  const void* srcHost,
+  size_t ByteCount,
+  CUstream hStream) {
+  cuda_api_enter_callback(NULL, gpu_placeholder_type_copyin);
+  CUresult result = auditor_exports->cuMemcpyHtoDAsync(dstDevice, srcHost, ByteCount, hStream);
+  cuda_api_exit_callback(NULL, gpu_placeholder_type_copyin);
+  return result;
+}
+
```
blue42u: Dead code. Remove unless needed later.

src/tool/hpcrun/sample-sources/nvidia.c
```diff
@@ -2,8 +2,9 @@
 
 // * BeginRiceCopyright *****************************************************
 //
-// $HeadURL: https://outreach.scidac.gov/svn/hpctoolkit/trunk/src/tool/hpcrun/sample-sources/papi.c $
-// $Id: papi.c 3328 2010-1/2-23 23:39:09Z tallent $
+// $HeadURL:
+// https://outreach.scidac.gov/svn/hpctoolkit/trunk/src/tool/hpcrun/sample-sources/nvidia.c $ $Id:
+// nvidia.c 3328 2010-1/2-23 23:39:09Z tallent $
```
blue42u: (Whole file): Not sure if there are consequential changes. Please minimize.

src/tool/hpcrun/threadmgr.c
```diff
@@ -376,7 +376,7 @@ hpcrun_threadMgr_data_put( epoch_t *epoch, thread_data_t *data, int no_separator
 
   // step 1: get the dummy node that marks the end of the thread trace
 
-  if (!no_separator) {
+  if (add_separator) {
```
blue42u: Unrelated change? Remove or rationalize.

src/tool/hpcstruct/Args.cpp
```diff
@@ -89,7 +89,7 @@ using std::string;
 //***************************************************************************
 
 // Size in bytes for parallel analysis of binaries
-#define DEFAULT_PSIZE     100000000   // 100MB
+#define DEFAULT_PSIZE     10000000   // 10MB
```
blue42u: Somewhat unrelated change. Factor out into a separate PR.

src/lib/prof-lean/hpcrun-fmt.h
```diff
@@ -171,7 +171,11 @@ hpcrun_fmt_hdr_free(hpcrun_fmt_hdr_t* hdr, hpcfmt_free_fn dealloc);
 #define HPCRUN_FMT_METRIC_SHOW_EXCLUSIVE  3
 #define HPCRUN_FMT_METRIC_INVISIBLE       4
 
-
+#define HPCRUN_FMT_DUMMY_NODE 65534
```
blue42u, line 174: Unrelated change?

src/lib/prof-lean/hpcrun-fmt.h
```diff
@@ -171,7 +171,11 @@ hpcrun_fmt_hdr_free(hpcrun_fmt_hdr_t* hdr, hpcfmt_free_fn dealloc);
...
+#define HPCRUN_FMT_GPU_RANGE_ROOT_NODE 65533
+#define HPCRUN_FMT_GPU_RANGE_NODE 65532
+#define HPCRUN_FMT_GPU_CONTEXT_NODE 65531
+#define HPCRUN_FMT_GPU_STREAM_NODE 65530
```
blue42u, line 178: With these additions, please remove the similar `#defines` from `lib/profile/hpcrun4.cpp`.

src/tool/hpcrun/main.c
blue42u, line 856: Unrelated change? Remove or rationalize.

This is generic code, how does this affect CPU threads?

src/tool/hpcrun/main.c
blue42u, line 1868: Dead code, remove.

src/tool/hpcrun/main.c
blue42u, line 1883: Code in comments, remove.

blue42u (changes requested):
Cursory read through the patch, looking for merge errors or other complications.

Also, I would really rather not have the hacky Hatchet support merged into `develop` if feasible.

src/tool/hpcrun/cct/cct.c
Jokeren: I'll take these dummy nodes related changes into another pr.

src/tool/hpcrun/control-knob.c
blue42u: AFAICT, the `else` clause is identical to the implementation of `strdup`. What's the fix?

src/tool/hpcrun/control-knob.c
Jokeren: @blue42u Can you explain why this isn't correct?

blue42u: Ignore that, I misremembered how `strtok` works. Looks fine.

src/tool/hpcrun/control-knob.c
Jokeren: Oh yes, thanks for pointing it out! I was referring to other code in this file but they were merged together.

Jokeren:
@blue42u Let me draft several additional prs to minimize files changes in this one. It is a pain to directly work on this large PR.

src/lib/profile/scope.cpp
Jokeren: @blue42u Yes, there's a compilation error `setw is not a member of std after taking it out` after taking out `#include<iomainip>`

Jokeren: gcc 8.5.0 was used

blue42u: @Jokeren I just tried in Spack, `hpctoolkit@develop %gcc@8.5.0` compiles just fine with the Spack-built GCC. I only just noticed, but there's literally an `#include <iomanip>` about 3 lines above your addition.

So remove.

src/tool/hpcrun/main.c
Jokeren: #521

src/tool/hpcrun/gpu/gpu-operation-item.h
```diff
@@ -81,8 +81,7 @@ typedef struct gpu_operation_item_t {
   s_element_t next;
   gpu_activity_channel_t *channel;
   gpu_activity_t activity;
-  atomic_int *pending_operations;
-  atomic_bool *flush;
```
Jokeren, line 85: `flush` is not used anywhere, can be safely deleted.

src/tool/hpcrun/gpu/gpu-op-placeholders.c
```diff
@@ -140,6 +140,18 @@ gpu_op_ccts_get
 }
 
 
+void
+gpu_op_ccts_update
```
blue42u, line 144: AFAICT unused function.

src/tool/hpcrun/gpu/gpu-op-placeholders.h
```diff
@@ -129,6 +130,15 @@ gpu_op_ccts_get
 );
 
 
+void
+gpu_op_ccts_update
```
blue42u, line 134: AFAICT unused function.

src/tool/hpcrun/gpu/nvidia/cupti-activity-translate.c
```diff
@@ -456,6 +456,7 @@ convert_branch
 }
 
 
+__attribute__((unused))
```
blue42u, line 459: This function is only "used" in a comment in `cupti_activity_translate`. If it's important to keep this code around for later, I would recommend an `#if 0` and a comment explaining why we're keeping dead code around like this.

And then revise `cupti_activity_translate` accordingly to minimize/remove the comment there and handle `CUPTI_ACTIVITY_KIND_SYNCHRONIZATION` as a proper case.

src/tool/hpcrun/gpu/nvidia/cupti-api.c
```diff
@@ -570,29 +601,41 @@ cuda_path
...
   const char *path = "libcupti.so";
-  int resolved = 0;
 
   static char buffer[PATH_MAX];
   buffer[0] = 0;
 
+#ifdef NEW_CUPTI
+  // XXX(Keren): Don't use the default cupti library under CUDA_HOME because
+  // since NVIDIA delivers us independent cupti libraries that contain bug
+  // fixes.
```
blue42u, line 630: Wouldn't this cause issues between mismatching CUDA and CUPTI? If it doesn't, why do we mess about in the old-CUPTI case with finding the CUDA install path at all?

blue42u (changes requested):
Another cursory inspection. Looking better now that its been minimized a bit.

Also a handful of changes that were extracted are still showing up in this PR, maybe rebase when the code stabilizes again.

src/lib/prof-lean/hpcrun-fmt.h
blue42u: This most definitely hasn't been resolved yet.

src/tool/hpcrun/main.c
blue42u: So, #521 changed the name and meaning of the argument, but didn't include this change which changes the usage.

configure.ac
```diff
@@ -4846,6 +4846,48 @@ AC_DEFINE_UNQUOTED([CUDA_INSTALL_PREFIX],
...
+# enable-new-cupti-analysis
+#-------------------------------------------------
+
+OPT_ENABLE_NEW_CUPTI_ANALYSIS=no
+
+AC_MSG_CHECKING([for new-cupti-analysis])
+
+AC_ARG_ENABLE([new_cupti_analysis],
+  AS_HELP_STRING([--enable-new-cupti-analysis],
+                 [Dump ccts for analysis]),
```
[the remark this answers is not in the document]
Jokeren: It also measures unwinding time, which could add extra overhead. I'll change it to a control knob.

src/tool/hpcrun/gpu/nvidia/cupti-api.c
Jokeren: 1. It does not cause issuing between CUDA and CUPTI. CUPTI > 11.0 can be used with any CUDA > 11.0. 
2. It is put here for now just to test nvidia's CUPTI patches. In principal, the correct CUPTI library should come with the CUDA library. However, we are using standalone CUPTI libraries just for testing. In deployment, we don't link CUPTI library outside of the CUDA directory.

src/tool/hpcrun/cct/cct.c
```diff
@@ -631,12 +631,30 @@ hpcrun_cct_insert_addr(cct_node_t* node, cct_addr_t* frm, bool unwound)
...
+  return context;
+}
+
 cct_node_t*
 hpcrun_cct_insert_dummy(cct_node_t* node, uint16_t lm_ip)
 {
   ip_normalized_t ip = { .lm_id = HPCRUN_DUMMY_NODE, .lm_ip = lm_ip };
   cct_addr_t frm = { .ip_norm = ip };
-  cct_node_t *dummy = hpcrun_cct_insert_addr(node, &frm, true);
+  cct_node_t *dummy = hpcrun_cct_insert_addr(node, &frm, false);
```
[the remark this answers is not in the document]
Jokeren: I'll revert it. It doesn't matter since dummy nodes are used at all.

src/tool/hpcrun/gpu/gpu-activity-process.c
```diff
@@ -455,6 +455,7 @@ gpu_kernel_block_process
 }
 
 
+__attribute__((unused))
```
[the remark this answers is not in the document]
Jokeren: Xiaozhu has explanations in nvidia.c. I'll just revert all changes in this file.

src/tool/hpcrun/gpu/gpu-correlation-id.h
```diff
@@ -52,7 +52,11 @@
 
 #include <stdint.h>
 
+#define GPU_CORRELATION_ID_START 0x8000000000000001
```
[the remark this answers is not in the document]
Jokeren: There are other correlation ids, such as OMPT correlation ids, which start from 0. Since 0x8000000000000001 has been used from a long time ago, I'd rather keep the same value.

src/tool/hpcrun/main.c
Jokeren: Good catch. Just fixed it.

src/lib/prof-lean/hpcrun-fmt.h
Jokeren: @blue42u You said **remove** similar defines from `lib/profile/hpcrun4.cpp`. I don't think hpcrun4.cpp still has the same definitions now.

blue42u: You are right, I was looking at the wrong branch when checking this. Sorry for the noise.

src/tool/hpcrun/gpu/nvidia/cupti-api.c
blue42u: > CUPTI > 11.0 can be used with any CUDA > 11.0.

Is that true in practice? I vaguely remember a case where we couldn't get any data out because the two had a version mismatch within the 11.x series.

If it is compatible, I would think it easier to just always use the one we link with and error if the loaded CUDA isn't a compatible version (which AFAIK shouldn't happen for Spack-built Toolkit since it messes with the RPATHs... which is also a complicating factor here). That way we don't have to worry about the headers changing in backward-incompatible ways between the major version series (because honestly I don't trust NVIDIA not to botch it somehow).

Either way, I would really prefer a different activation conditional here, having it bound to `--enable-new-cupti` doesn't make sense to me yet. Maybe an environment var/control knob/`-dd` to skip the search for the CUPTI bundled with CUDA?

src/tool/hpcrun/gpu/gpu-activity-process.c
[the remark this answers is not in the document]
blue42u: That would add the warning about an unused function back though, wouldn't it? I'm fine with `#if 0`'ing the whole block if that's the right approach.

Are SYNCHRONIZATION events supposed to be suppressed just for NVIDIA? Because AFAIK this is generic code, so if its a vendor-specific suppression it should be in `nvidia/*.c` somewhere. Like `cupti-activity-translate.c` with my very similar comment.

src/tool/hpcrun/gpu/nvidia/cupti-api.c
Jokeren: > Is that true in practice? I vaguely remember a case where we couldn't get any data out because the two had a version mismatch within the 11.x series.

Well, this is how CUPTI is designed to be. After CUDA 11.0, these CUDA libraries should have good forward compatibility and backward compatibility among major versions (11.x). I admit that there were still minor issues though.

> Either way, I would really prefer a different activation conditional here, having it bound to --enable-new-cupti doesn't make sense to me yet. Maybe an environment var/control knob/-dd to skip the search for the CUPTI bundled with CUDA?

Let me rephrase my sentences here. I don't want this special cupti binding at all, and it makes no sense to link a CUPTI library with a CUDA library with different versions. It is put here just because CUPTI is buggy at this moment so that we cannot use the default CUPTI library shipped with CUDA. 

And the whole point of using NEW_CUPTI is to have a temporary version that tests the new pc sampling api. Once we could confirm that CUPTI does not have significant problems, I'll remove the NEW_CUPTI macros and merge similar code.

src/tool/hpcrun/gpu/gpu-activity-process.c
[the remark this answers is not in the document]
Jokeren: > That would add the warning about an unused function back though, wouldn't it? I'm fine with #if 0'ing the whole block if that's the right approach.

Yes, the compiler outputs a warning message, and the warning message is also in the master branch. Synchronization events are supposed to be used for all GPUs. 

From my perspective, I like your idea to minimize code differences between two branches. So it is probably better to eliminate the differences regarding the synchronization activity in this pull request. We can create a github issue to discuss this problem and cleanup related code.

blue42u:
Stack pointer for any arch can be (roughly) gotten via `__builtin_frame_address(1)` (or so): https://gcc.gnu.org/onlinedocs/gcc/Return-Address.html

blue42u:
[This repository has moved to GitLab, please continue this PR there.](https://gitlab.com/hpctoolkit/hpctoolkit/-/issues/519)

blue42u:
The previous link was incorrect, apologies. The corrected link is [here.](https://gitlab.com/hpctoolkit/hpctoolkit/-/merge_requests/519)

PR #526, opened 2022-03-08
GPU kernel level profile view improvement for Intel Level0 

mxz297:
This PR ensures that Level 0 support for GPU kernel level measurement works with latest level 0 version and intel gpu runtime. It also uses Level 0 to dump a Intel GPU debug info binary and links source line info with gpu kernels.


src/tool/hpcrun/gpu/amd/rocm-binary-processing.c
```diff
@@ -46,12 +46,11 @@
 //******************************************************************************
 
 #include <gelf.h>
-#include <errno.h>     // errno
-#include <fcntl.h>     // open
-#include <sys/stat.h>  // mkdir
+#include <fcntl.h>
+#include <sys/types.h>
```
jmellorcrummey, line 50: sys/types is twice

mxz297: @jmellorcrummey done

PR #527, opened 2022-03-08
Openssl option

Paul-Ferrell:
Added the option to specify an openssl library instead of mbedtls, as on many systems openssl is simply available (plus the many other reasons for this detailed in emails).

This adds a `--with-openssl=<openssl_path>` option to configure, which is only used if mbedtls is not provided.


Paul-Ferrell:
I still need to test these changes (they build seemingly correctly) by building with Spack and running the built-in unit tests.

jmellorcrummey:
@mwkrentel i wanted to make sure that you are aware of this.

@Paul-Ferrell does LANL want to build HPCToolkit with spack or manually with configure?

@mwkrentel it seems we could accommodate this in spack with a +openssl option that would swap mbedtls for openssl.

Paul-Ferrell:
Looking through the changes, it looks like autotools made some unexpected changes to several additional files, like `configure.guess`. I'm going to push a clean branch of this with just the changes to `configure.ac` and `crypto-hash.c`, that way you guys can just autogen all the files in your typical environment so it doesn't create such a mess.

Paul-Ferrell:
Closed in favor of #528.

PR #528, opened 2022-03-09
Openssl option

Paul-Ferrell:
This has just my direct changes, without the mess autogen makes. Autogen will still need to be run to fix the configure script and Makefile.in files.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -1862,6 +1862,7 @@ default_libunwind_prefix=no
 default_lzma_prefix=no
 default_mbedtls_prefix=no
 default_memkind_prefix=no
+default_openssl_prefix=no
 default_papi_prefix=no
 default_perfmon_prefix=no
 default_tbb_prefix=no
@@ -1977,6 +1978,9 @@ then
   fi
   if find_spack_pkg memkind ; then
     default_memkind_prefix="$spack_ans"
+  fi
+  if find_spack_pkg openssl ; then
+    default_openssl_prefix="$spack_ans"
   fi
   if find_spack_pkg papi ; then
     default_papi_prefix="$spack_ans"
@@ -3004,9 +3008,9 @@ AM_CONDITIONAL(OPT_LZMA_STATIC, [test "$lzma_static" = yes])
 # Note: we require libmbedcrypto.a to be compiled with -fPIC.
 
 MBEDTLS="$default_mbedtls_prefix"
-MBEDTLS_IFLAGS=
-MBEDTLS_LIBS=
-MBEDTLS_LIB=
+CRYPTO_IFLAGS=
+CRYPTO_LIBS=
+CRYPTO_LIB=
 
 AC_ARG_WITH([mbedtls],
   [AS_HELP_STRING([--with-mbedtls=PATH],
@@ -3021,15 +3025,15 @@ case "$MBEDTLS" in
     if test ! -f "${MBEDTLS}/include/mbedtls/md5.h" ; then
       AC_MSG_ERROR([unable to find md5.h in: $MBEDTLS])
     fi
-    MBEDTLS_IFLAGS="-I ${MBEDTLS}/include"
+    CRYPTO_IFLAGS="-I ${MBEDTLS}/include -D MBEDTLS_CRYPTO"
     for lib in $multilib_path ; do
-      if test -f "${MBEDTLS}/lib/libmbedcrypto.a" ; then
-	MBEDTLS_LIB="${MBEDTLS}/lib"
-        MBEDTLS_LIBS="${MBEDTLS}/lib/libmbedcrypto.a"
-	break
+      if test -f "${MBEDTLS}/${lib}/libmbedcrypto.a" ; then
+    	CRYPTO_LIB="${MBEDTLS}/${lib}"
+        CRYPTO_LIBS="${MBEDTLS}/${lib}/libmbedcrypto.a"
+	    break
       fi
     done
-    if test "x$MBEDTLS_LIBS" = x ; then
+    if test "x$CRYPTO_LIBS" = x ; then
       AC_MSG_ERROR([unable to find libmbedcrypto.a in: $MBEDTLS])
     fi
     num_pkgs=`expr 1 + $num_pkgs`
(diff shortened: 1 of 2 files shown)


mwkrentel:
@Paul-Ferrell Yes, sorry about the autotools changes.  We use autotools in
"maintainer mode".  The plus side is that users don't need to build
autotools to build hpctoolkit, the down side is that it adds more work
to developers.  This all made more sense pre-spack.

If you were to build exactly autoconf 2.69, automake 1.15.1 and
libtool 2.4.6, then you would see a more proportionate diff.
We keep a script to do this for us.

Anyway, it looks like you've tried to follow the flow of the existing
code, so thank you.  But I do have a few style comments.

1. In configure.ac, we normally use 2 spaces for indenting, there are
some places where you've used 4.

2. Also in configure.ac, I'd prefer never to use [ or [[ except as
quotes.  This is an m4 thing.  Autoconf uses [ and ] as the quote
characters and it's easy to introduce a mistake.

Instead use 'test' as in: if test "$MBEDTLS" = no ; then ....

3. In crypto-hash.c, I'd prefer to write MD5_FUNC in all caps to make
it look like a macro.

More importantly, for mbedtls, we use libmbedcrypto.a, static library
compiled with -fPIC and we test for this in configure.ac.  We'll need
to do the same for openssl, that is use static, not shared library.
(Your patch has a typo of libopenss.so instead of libcrypto, but it needs
to be static anyway.)

And we need to check if there's an option to add -fPIC to openssl in
spack.  Right now, I don't see one, unless by chance it always builds
that way.

This is part of how we hide 3rd-party library names so that we don't
clash with an app that also uses a different version of the same
library.

mwkrentel:
@Paul-Ferrell Btw, I guess I still don't see the security hole.  If an app wanted to
implement an internal hash table and use md5 as the hash function,
where's the security hole?  (That's what we do.)

So, you're going to continue to use spack?
Are you going to use a modified spack that bans all md5 implementations
execpt one?

Paul-Ferrell:
### On your code quality comments  
I'm 100% on-board with all of them.

### So there's two security issues, the one I spotted in the code and the concern about mbedtls/md5.
#### On the code one 
So let's say I made a hash table where I decided I only needed 8 bytes of hash. I could call `crypto_hash_compute` with a 8 byte buffer and the buffer length set to 8, and reasonably expect the function to do the right thing. Except it doesn't do the right thing. Unlike `strncopy` or most any other function that takes a buffer length, it always writes 16 bytes to the buffer (but only zeros out the first 8). The only check is that you don't ask for too many bytes of hash. 

#### On the more general one regarding why this is all necessary at all:
MD5 is not the issue. There's zero threat there. 

The problem is that we aren't building hpctoolkit as a one-off thing - we're building software environments (that include hpctoolkit) that will be around for the entire life of the system. An executable built on day 1 of the system's life has to still run on the last day (and this frequently the case). We're using Spack to build all this out now. New stuff gets piled on top, but the old bits never go away. Including old, vulnerable implementations of security critical libraries like openssl, gnuTLS, and mbedTLS. 

Even though those probably aren't a huge risk given their use on our systems, we still have to update them by policy. Which is a pain, and not always possible! OS vendors stick with a set version for an OS release, and just backport fixes to that version. So even if openssl 1.0.2 (in source form) never gets super critical updates in the future, RHEL's openssl-1.0.2 package will. So we would have to upgrade the openssl version (and potentially break binary comparability) or backport those fixes ourselves. OR... We could just use the well-tested vendor provided packages that do get those updates and backports. 

Except... mbedTLS isn't provided by most vendors. So do we make an exception? Do we massively modify the mbedTLS Spack package so it only builds MD5 (and other non-issues bits) as a variant? Do I make a simple ask of the hpctoolkit devs and add an option to use mbedtls or another library we're already pulling from the system? I looked at your code and both the mbedTLS and openssl documentation, and saw that this was both the easiest route and the cleanest in the long run.

jmellorcrummey:
@Paul-Ferrell 

1. I understand your security concerns and we'll accept your change once we've developed and tested the compatible spack support
2. As for your code concerns, I am aware that the crpyto hash API we expose is prone to errors because you have to provide the appropriate hash length. I wrote a wrapper that was more convenient when hashing gpu binaries on disk. 

https://github.com/HPCToolkit/hpctoolkit/blob/master/src/lib/prof-lean/elf-hash.c

This API is simple and has no problematic usage. It takes in a file name and returns an allocated string of the appropriate length. (The usage is not problematic as long as you free the character string when you are done with it :-).

I expect that we will convert to a similar interface eventually for the crypto hash function that we expose to avoid hash length request and buffer length errors.

mwkrentel:
@Paul-Ferrell I think I mostly understand.  Sorry for being dense.

The more serious problem is the adjustments to openssl so we can
hide the library inside hpcrun so it doesn't clash with the app.

jmellorcrummey:
@Paul-Ferrell We're closing this pull request since @mwkrentel addressed your concern by incorporating an implementation of MD5 inside HPCToolkit. We no longer need to an MD5 implementation from OpenSSL.

It is worth noting that the public domain implementation of MD5 that you pointed us to crystalized why we used an externally vetted implementation in MBEDTLS. We had to fix the public domain implementation: it dumped core (it overwrote the return address) on anything larger than a few hundred bytes. You get what you pay for with free software :-).

Paul-Ferrell:
Sounds good.


________________________________
From: John Mellor-Crummey ***@***.***>
Sent: Wednesday, April 6, 2022 8:32 AM
To: HPCToolkit/hpctoolkit
Cc: Ferrell, Paul Steven; Mention
Subject: [EXTERNAL] Re: [HPCToolkit/hpctoolkit] Openssl option (PR #528)


@Paul-Ferrell<https://urldefense.com/v3/__https://github.com/Paul-Ferrell__;!!Bt8fGhp8LhKGRg!QwlztMHpT-NRmbVV-JMBSnZz23lXgV2Nru9fcPUlrp8vmIQbaZQv1SErfDbTxJUp$> We're closing this pull request since @mwkrentel<https://urldefense.com/v3/__https://github.com/mwkrentel__;!!Bt8fGhp8LhKGRg!QwlztMHpT-NRmbVV-JMBSnZz23lXgV2Nru9fcPUlrp8vmIQbaZQv1SErfLYMmdTt$> addressed your concern by incorporating an implementation of MD5 inside HPCToolkit. We no longer need to an MD5 implementation from OpenSSL.

It is worth noting that the public domain implementation of MD5 that you pointed us to crystalized why we used an externally vetted implementation in MBEDTLS. We had to fix the public domain implementation: it dumped core (it overwrote the return address) on anything larger than a few hundred bytes. You get what you pay for with free software :-).

PR #529, opened 2022-03-10
Fix: Separate libiberty fails to compile

blue42u:
When compiling with `--with-spack`, if the Binutils it picks up was compiled with `~libiberty` (which is default), the build fails with the error:
```
In file included from demangle.c:59:
.../hpctoolkit/src/include/gnu_demangle.h:68:10: fatal error: libiberty/demangle.h: No such file or directory
   68 | #include <libiberty/demangle.h>
      |          ^~~~~~~~~~~~~~~~~~~~~~
compilation terminated.
```

This patch adds includes handling to the separated libiberty logic.

### DIFF
--- a/configure
+++ b/configure
@@ -21116,6 +21116,13 @@ $as_echo_n "checking for libiberty.a for binutils... " >&6; }
   found_libiberty=no
   case "$LIBIBERTY" in
     /* )
+      # demangle.h must exist
+      if test -f "${LIBIBERTY}/include/libiberty/demangle.h" ; then
+        BINUTILS_IFLAGS="${BINUTILS_IFLAGS} -I${LIBIBERTY}/include"
+      else
+        as_fn_error $? "unable to find libiberty/demangle.h in: $LIBIBERTY" "$LINENO" 5
+      fi
+
       for lib in $multilib_path ; do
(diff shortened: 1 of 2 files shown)


mwkrentel:
LGTM, I approve

I remember some 6-9 months ago John ran into this problem and I told
him I'd fix it.  But then, a couple weeks later, I asked him about it,
but neither of us could remember what the issue was.  :-(

mwkrentel (approved):

jmellorcrummey:
I have a question about our spack dependences. we only depend upon binutils. dyninst depends on libiberty. is libiberty only built because dyninst demands it and we choose to use it if it happens to be there? shouldn't we add libiberty as an hpctoolkit spack dependence?

mwkrentel:
@jmellorcrummey We directly demangle names in a couple of places in our code.

Toolkit is setup to allow libiberty from  binutils or libiberty.
Currently, we tell spack to build binutils +libiberty which is one
way of covering our needs.  The libiberty backup is more for autotools
than for spack.

The real issue is the remaining use of binutils.  I **think** all we
really use are the variable types from binutils, but they're
EVERYWHERE.  If we cleaned that out, then we could use just libiberty.
Or, maybe we'd find some remaining piece that hopefully we could
satisfy in some other way.

PR #530, opened 2022-03-11
Fixing hpcrun_threadMgr_data_get problem

dejangrubisic:
This PR fixes the problem of overwriting thread_data structure by setting mem_pool_initialized on right place.


jmellorcrummey:
I debugged this with Dejan, so I am familiar with the solution.

PR #531, opened 2022-03-15
Apply CFLAGS to libhpcrun_{,fake_}audit.so

blue42u:
Fixes an issue where --enable-develop/debug didn't compile the auditor with debug flags, so debugging was more difficult than it should have been.

...But seriously, Autogoo is a mess, this sort of thing should be automatic.


jmellorcrummey (commented):
LGTM

mwkrentel:
> ...But seriously, Autogoo is a mess, this sort of thing should be automatic.

@blue42u  All this HPC_CFLAGS stuff is from the mind in Nathan,
it's not all autotools fault.
It's the kind of thing that I'd clean up if we were staying with autotools.

PR #532, opened 2022-03-16
Remove Gotcha as a dependency

blue42u:
It's currently unused, and likely will remain unused in favor of LD_AUDIT.

Requesting removal now since I've been running into Gotcha build failures on some platforms (e.g. Fedora 36).


jmellorcrummey (approved):
LGTM

mxz297:
@blue42u @jmellorcrummey  Will `spack install hpctoolkit@master` fail since the configure script no longer takes the --with-gotcha option?

blue42u:
@mxz297 It should work fine, for good or ill `./configure` ignores arguments it doesn't understand. So the most you get is a warning in the build log:
```
configure: WARNING: unrecognized options: --with-gotcha
configure: INFO: from user: CC='.../spack/lib/spack/env/gcc/gcc'  CFLAGS=''
configure: INFO: from user: CXX='.../spack/lib/spack/env/gcc/g++'  CXXFLAGS=''
...
```

mxz297:
@blue42u I see. Then that should be fine.

PR #535, opened 2022-03-20
remove assertion in hpcrun_get_threaded_data

jmellorcrummey:
the assertion is false when monitoring GNU bash, version
4.4.23(1)-release (x86_64-suse-linux-gnu), which appears to fork
a threaded process.

### DIFF
--- a/src/tool/hpcrun/thread_data.c
+++ b/src/tool/hpcrun/thread_data.c
@@ -230,7 +230,12 @@ hpcrun_threaded_data
   void
 )
 {
+#if 0
+  // this assertion isn't true for when a threaded process forks and the child then creates a thread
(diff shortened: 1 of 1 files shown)


mwkrentel:
Well, I looked at the code and read the comments (none) and I don't
really understand the logic of initializing the thread pointers for
complex cases like fork/pthread-create.

But it seems to me that if hpcrun doesn't handle the case of fork from
a threaded process correctly, then that's where I'd look.  It sounds
like hpcrun is not correctly reinitializing after fork.  Remember,
it's the thread that calls fork that becomes the sole, master thread
of the new process.  So, maybe hpcrun should reset the main thread to
that in the child of fork.

mwkrentel:
Or, maybe the code is correct and it's the assert that's confused about
the threaded fork case, I dont know.

PR #537, opened 2022-03-24
Merge pull request #497 from HPCToolkit/rocprofiler_support

mxz297:
This PR was originally merged to the develop branch and it is now cherry-picked to master so that we can do a release for AMD platforms. This seems to work reasonably well on Crusher using rocm-5.0.2 with the matrix multiplication test. Original PR message below:

This is a huge merge to develop . It includes three main components:

1. Preliminary rocprofiler support for getting AMD GPU hardware counters.
2. Prelminary support for AOMP (AMD's openmp implementation).
3. Breakdown of hpcrun initialization and reordering.

(cherry picked from commit 627d4da6efa9402a291183e2307ac4c39b9d5895)


src/lib/prof-lean/crypto-hash.h
```diff
@@ -159,6 +159,10 @@ crypto_hash_self_test
   int verbose
 );
 
+#if defined(__cplusplus)
```
jmellorcrummey, line 162: This if defined(__cplusplus) should be deleted. It is an unmatched duplicate inside an if 0. The whole if 0 can be deleted if you choose.

mxz297: @jmellorcrummey Done

configure.ac
```diff
@@ -5013,6 +5053,171 @@ AC_SUBST([OPT_GTPIN_IFLAGS])
 AC_SUBST([OPT_GTPIN_LIBDIR])
 
 
+#-------------------------------------------------
```
jmellorcrummey, line 5056: remove redundant copies of --with-igc, with-metrics-discovery, with-gtpin

jmellorcrummey (approved):
The PR is OK. There is a bit of redundant code (e.g. in configure.ac). There is some unused code (the gpu_monitor layer) and the hip_api layer. However, our plan is to clean this up in develop rather than worry about it here. Master will soon be replaced.

PR #538, opened 2022-03-24
Split printProc() into: earlyFormatProc() which performs most of the

mwkrentel:
format conversion (except index numbers and gaps) and can run in
parallel, and finalPrintProc() which runs serially and adds the index
numbers and writes to the struct file.

This allows the bulk of the format conversion to run in parallel and
dramatically reduces the "print engine spike" in hpcstruct.


mwkrentel:
John, in my office, says "Ok".

PR #539, opened 2022-03-25
Rewrite and fixup the auditor

blue42u:
The behavior should be nearly identical, but the code has been reviewed and cleaned up based on my refined understanding of LD_AUDIT and the issues surrounding it. In particular, the global state machine is replaced with more natural piecewise logic.

Notable changes:
 - The auditor_initialize hook has been removed, auditors are no longer able to trigger Run initialization. Run initialization already happens during static constructors (or from eg. pthread_create), this cleans up the loose end.
 - Added a new hook mainlib_disconnect. This allows libhpcrun.so to fully control when it will receive notifications: only after mainlib_connected and before mainlib_disconnect. (Fixes no bugs, but is cleaner.)
 - Additional fallbacks for loading program headers. This handles cases where the programs headers are not at the front of the file. (Fixes @Lucindaaaa's issues with strange libraries in Intel software.)
 - Prefer /proc/self/exe to AT_EXECFN for getting the main executable's path. (Fixes a bug where scripts with a shebang would end up with the script as the main executable load module.)

### DIFF
--- a/src/tool/hpcrun/audit/audit-api.h
+++ b/src/tool/hpcrun/audit/audit-api.h
@@ -65,9 +65,6 @@ typedef struct auditor_map_entry_t {
   // handling the main application name.
   char* path;
 
-  // Base address of the file. Points to the ELF header.
-  void* ehdr;
-
   // Byterange of the mapped region
   void* start;
   void* end;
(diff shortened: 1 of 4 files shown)


jmellorcrummey (commented):
I'd like you to talk me through the change. I don't understand the comment about the change to initialization.

jmellorcrummey (approved):
I discussed the changes with Jonathon. While I understand the high-level goals of eliminating the state machine and scrapping handling for an initialization callback, I didn't meticulously inspect and understand all of the details. I'm just trusting Jonathon on this one.

PR #545, opened 2022-03-29
Remove undefined call

mxz297:
Fix error below when there is a typo in the hpcrun metric name

```
 symbol lookup error: libhpcrun.so: undefined symbol: is_event_to_exclude
```


jmellorcrummey (approved):
LGTM

PR #546, opened 2022-03-29
don't connect to auditor in a forked child

jmellorcrummey:
have both the auditor and hpctoolkit ensure that auditor does not execute connect in forked child.


blue42u (approved):
LGTM

PR #553, opened 2022-04-05
Implement an interface to allow specifying a paired CPU & GPU timesta…

mxz297:
…mp to align CPU & GPU traces


jmellorcrummey (approved):
Discussed with Xiaozhu. Understood. LGTM as an interim patch.

PR #554, opened 2022-04-05
Fix doc

laksono:
Updates to sync with the latest version of hpctoolkit which include the manual, man-page and www pages.
Minor update: move Philip from current team to ex-member.


jmellorcrummey (approved):
I made a pass over the Release Notes and added some new material to the OpenMP chapter in the manual about GPUs. 

I think it looks good enough. I don't have the energy to copy-edit the whole manual line by line at present.

mwkrentel:
I think the release date should be 2022.04.15, not 05.

If it were me, I would paginate the lines, break them around 80 chars.
As it is, I have to slide back and forth in the window.

Other than that, I approve.

@laksono Thanks for doing this.

mwkrentel (approved):

PR #555, opened 2022-04-06
In prof-lean, replace mbedtls with a smaller, direct C implementation

mwkrentel:
of MD5 for crypto-hash (and fix one bug) from:

  https://github.com/galenguyer/md5

Remove all the references to mbedtls.  This means that we no longer
depend on mbedtls which LANL did not want to use.

Update the crypto-hash unit test for the new MD5.

----------

ping @jmellorcrummey I think this will do what LANL wants.
I've removed all trace of mbedtls.
I also updated your unit test in crypto-hash, if you want to try that.


mwkrentel:
Most of the diff is in generated files (Makefile.in).  The Makefile.am
files and configure.ac simply drop MBEDTLS, all the new code is from
lib-prof-lean.  That leaves just:

```
M       src/lib/prof-lean/crypto-hash.c
A       src/lib/prof-lean/md5.c
A       src/lib/prof-lean/md5.h
```

jmellorcrummey (approved):
LGTM.

PR #558, opened 2022-04-11
unwind dejan's papi changes

jmellorcrummey:
- make PAPI work on CPU
- eliminate ROCM support that never worked
- roll back changes to CUDA support which are problematic because they don't support forking

NOTE: papi CUDA support probably shouldn't be compiled in, though it is


jmellorcrummey:
@laksono I believe that this commit fixes all papi CPU issues

mxz297 (approved):
LGTM

PR #559, opened 2022-04-11
fix compiler warnings

jmellorcrummey:
fix issues that caused a large collection of compiler warnings

### DIFF
--- a/src/tool/hpcrun/files.c
+++ b/src/tool/hpcrun/files.c
@@ -149,7 +149,7 @@
 
 struct fileid {
   int  done;
-  long host;
+  unsigned int host;
   int  gen;
 };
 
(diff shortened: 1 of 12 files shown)


blue42u:
FYI with GCC 11.2 [there are still a number of warnings](https://github.com/HPCToolkit/hpctoolkit/files/8460858/Wall.log), and of course [even more with `-Wextra`](https://github.com/HPCToolkit/hpctoolkit/files/8460859/Wextra.log). The string-manipulation ones in particular don't seem completely innocuous.

jmellorcrummey:
@blue42u I never said I fixed all of the warnings :-). 
- I was distracted by the fact that PAPI was broken and shifted to fixing that. I had it down to two warnings in files.c about the FILE_TEMPLATE and the two for printing the debug output for binary trees. I actually don't know how to fix the issues with the FILE_TEMPLATE. perhaps the gcc 11 warnings are more helpful.
- gcc 11 shows a lot more. fixing the new ones in your log is a good idea. 
- ignore the PAPI warnings. the papi code gets replaced by another PR.

jmellorcrummey:
merging to eliminate these warnings before working on master with gcc 11.2.0 on crusher

Issue #556, opened 2022-04-06
hpcrun yields Segmentation fault depending on the event order


xeon-j:
`-e BLOCKTIME -e gpu=nvidia` works fine but `-e gpu=nvidia -e BLOCKTIME` causes segmentation fault at the end of the run.

Output from failed run (LAMMPS + Kokkos CUDA):

```bash
mpiexec -n 1 \
    hpcrun -e BLOCKTIME -e gpu=nvidia \
    ./lmp ...

[FPGA09:34461] *** Process received signal ***
[FPGA09:34461] Signal: Segmentation fault (11)
[FPGA09:34461] Signal code:  (128)
[FPGA09:34461] Failing at address: (nil)
[FPGA09:34461] [ 0] /opt/hpctoolkit/lib/hpctoolkit/ext-libs/libmonitor.so(+0x77a9)[0x7f83df3747a9]
[FPGA09:34461] [ 1] /lib/x86_64-linux-gnu/libpthread.so.0(+0x143c0)[0x7f83def3f3c0]
[FPGA09:34461] [ 2] /opt/hpctoolkit/lib/hpctoolkit/libhpcrun.so(hpcrun_get_num_metrics+0x40)[0x7f83df3c9be0]
[FPGA09:34461] [ 3] /opt/hpctoolkit/lib/hpctoolkit/libhpcrun.so(hpcrun_new_metric_data_list+0x37)[0x7f83df3ca1e7]
[FPGA09:34461] [ 4] /opt/hpctoolkit/lib/hpctoolkit/libhpcrun.so(hpcrun_metric_set_loc+0x5f)[0x7f83df3ca2af]
[FPGA09:34461] [ 5] /opt/hpctoolkit/lib/hpctoolkit/libhpcrun.so(hpcrun_metric_std+0x31)[0x7f83df3ca301]
[FPGA09:34461] [ 6] /opt/hpctoolkit/lib/hpctoolkit/libhpcrun.so(+0x22699)[0x7f83df3bb699]
[FPGA09:34461] [ 7] /opt/hpctoolkit/lib/hpctoolkit/libhpcrun.so(+0x20c8b)[0x7f83df3b9c8b]
[FPGA09:34461] [ 8] /opt/hpctoolkit/lib/hpctoolkit/libhpcrun.so(+0x1b365)[0x7f83df3b4365]
[...]

PR #560, opened 2022-04-12
Fix blocktime

laksono:
Minor update to fix issue #556 :
- Remove access to perf info for debugging purpose.
- Use blocktime metric kind to separate with other metric kinds


src/tool/hpcrun/sample-sources/perf/kernel_blocking.c
```diff
@@ -126,9 +128,11 @@ blame_kernel_time(event_thread_t *current_event, cct_node_t *cct_kernel,
   // ----------------------------------------------------------------
   // it's important to always count the number of samples for debugging purpose
   // ----------------------------------------------------------------
-
-  thread_data_t *td = hpcrun_get_thread_data();
-  td->core_profile_trace_data.perf_event_info[metric_blocking_index].num_samples++;
+// fix issue #556: since we use blocktime metric kind, we don't have access to perf info
```
Jokeren, line 131: So perf's metrics kind cannot be used by blocktime? Sorry I don't entirely understand the comment.

laksono: @jmellorcrummey The first commit is wrong. I thought it was the solution.
@Jokeren The number of samples is not the same as the metric value.  Sometimes I need to know the number of samples to check the correctness of the metric value. 
The blocktime event is a "derived" event of perf event which uses perf's context switch event. It shouldn't have access to any perf information.

jmellorcrummey:
fixing the blocktime and context switch metrics involved reorganizing the implementation of metrics managed by Linux perf_events. After the reorganization, we observe that sometimes the blocktime and context switch metrics are correct. In some cases, when the exclusive time maps to samples within the kernel (with kernel call paths appended below user call paths), the exclusive metrics can be lost. we have decided to accept the PR now and defer full resolution of this issue to prof2.

PR #563, opened 2022-04-17
fix issues with renaming structure files in the cache

jmellorcrummey:
- replace "@echo" with "echo" in renamestruct.sh script
  (@echo is a make construct but not legal bash)

- fix monitor_execve failure warning when using hpcrun to measure
  hpcstruct. the issue was that hpcrun was trying to exec
  a script directly. instead of running the rename script
  with command  "<path>/renamestruct.sh", run it with command
  "/bin/sh <path>/renamestruct.sh". that way, /bin/sh can be
  exec'ed. i think that if such an exec failed in the shell,
  the shell probably figures out how to launch the command
  by finding the "#!/bin/sh" in the first line of the shell
  script.

### DIFF
--- a/src/tool/hpcstruct/SingleBin.cpp
+++ b/src/tool/hpcstruct/SingleBin.cpp
@@ -290,7 +290,7 @@ doSingleBinary
   // if a cache is in use, ensure that the module path in the new .struct file is correct.
   //
   if (use_cache == true ) {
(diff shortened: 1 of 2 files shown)


jmellorcrummey:
@martyitz I'm tagging you so you see these adjustments. I am not asking for a review. I've tested the fixes myself.

PR #507, opened 2022-01-19
Add cache for hpcstruct

jmellorcrummey:
Cache binary analysis results of CPU and GPU binaries.

Currently, the cache can record hpcstruct files and gaps files. Some refactoring of the API will enable it to hold other kinds of analysis results, e.g. dataflow analysis results for GPU binaries.

PR #570, opened 2022-04-27
Remove Struct's reliance on scripts + Proftt

blue42u:
Currently `hpcstruct` invokes shell scripts to implement brand new use cases. Measurement directory support (#421) runs a Makefile, which in turn invokes `hpcproftt` to parse `*.hpcrun` files. Cache support (#507) runs `renamestruct.sh` to adjust the binary name when pulling from the cache.

While these approaches "work," they are fraught with design concerns:
 - Most critically, **shells do not handle errors gracefully**. Subcommand failures are often ignored (eg. in a pipeline). This is generally very difficult to fix, and failing to handle errors results in confusing logs or worse, unusable output. These are critical issues for a user-facing tool like `hpcstruct`.
 - `hpcproftt` is primarily a development tool, while it is relatively simple it was not written to be fast or robust. Relying on it in a user-facing tool like `hpcstruct`, or in general requiring it as part of a complete install, does not match it's original purpose.
 - `/bin/sh` is not a stable target, on Red Hat it's Bash but on Debian it's Dash (a minimal POSIX-compliant shell). Bash-isms sometimes leak in (like #429) and break the tool in mysterious ways (when running on Debian like me).

Also, using scripts that need to be installed alongside `hpcstruct` complicates running uninstalled (out of the build tree). While this case isn't supported by the current Autotools build system, it is by the Meson port (#320). (Handling this on the side makes my life easier updating the port to the latest `develop`.)

See the commit log for details on the new approach.

Current known issues:
 - ~~Control-C on `hpcstruct --gpucfg yes x.gpubin` does not stop, because `system()` blocks `SIGINT` in the parent (see `man system`). With this PR, this causes Control-C on `hpcstruct --gpucfg yes measurements_dir` to only stop the "overseer" process and not the "worker" processes, orphaning the last batch of worker processes. How best to solve one or both of these issues requires some investigation.~~

   **UPDATE:** Control-C will now wait for the current batch of workers to complete. If `--gpucfg yes` is used, this may not be immediate and may cause unnecessary NVIDIA warnings.
 - ~~`*.warnings` files are no longer generated, the errors from workers are output directly to the shared stderr. This may cause inconvenience when `nvdisasm` spews thousands of lines of error. Whether this should be handled here or as part of the `nvdisasm` invocation requires discussion.~~

   **UPDATE:** `*.warnings` files are now generated.


blue42u:
In response to @jmellorcrummey:
> the reason for the warnings files is that we have seen nvdisasm spew 100K lines of errors. having that appear on stdout is not useful.

The same issue is present when parsing a single binary (although it's a far less common use case), so IMHO it's the error/output handling around `nvdisasm` at fault here.

Maybe we could redirect its output to a `--log-file` (on request) and leave a (short) warning in stderr, but I'd rather wait until #522 is merged before messing with that code. (Ping @Jokeren)

> making complete result files appear atomically and respecting results files that are present (with the exception of when the —gpucfg argument to hpcstruct was changed) lets one control-c hpcstruct and restart it without throwing away what it may have already computed.

IMHO the cache has the same effect, and the implementation is far more appropriate (no sensitivity to timestamps, no auxiliary not-quite-temporary files, proper cleanup of outputs on crash...). The output is also clearer IMO since it lists all the binaries every time.

We already recommend using a cache, but maybe we could add a cache within the measurements directory by default?

jmellorcrummey:
> The same issue is present when parsing a single binary (although it's a far less common use case), so IMHO it's the error/output handling around nvdisasm at fault here.

When the errors are due to a single binary, the answer is obvious: restart and put the output in a log file. When the errors spew from analysis of a measurement directory, you provide no choice but to restart analysis of everything and then hunt through the mass of output to see if you can figure out which file is the culprit. Also, from what your comments describe (everyone uses stdout/stderr), the errors from analysis of multiple files are likely to be interleaved (and not necessarily at the line level either). For that reason, when analyzing multiple files, dumping the output in a warnings file is much better. 

It used to be the case that the Makefile would create a warnings file for every file analyzed and then remove any warnings file that was empty after analysis was complete. Marty's changes to use them to record cache state made them always non-empty, which was a downgrade.

Jokeren:
https://github.com/HPCToolkit/hpctoolkit/pull/522 does not report any problem with merging into `develop`. I am not sure what test cases are necessary though before merging. Maybe I should contact Marty?

blue42u:
`*.warnings` files are now generated as before, except the exit code is used to pass the cache state back to the "overseer" so there should only be `*.warnings` when there are real warnings. This should resolve @jmellorcrummey's concerns.

I've also added some code to handle `^C` as best as possible without adjusting `nvdisasm` invocation, which handles the other issue I knew of.

If there are no other issues, ready for review.

blue42u:
Updated and rebased on top of #598

blue42u:
[This repository has moved to GitLab, please continue this PR there.](https://gitlab.com/hpctoolkit/hpctoolkit/-/issues/570)

blue42u:
The previous link was incorrect, apologies. The corrected link is [here.](https://gitlab.com/hpctoolkit/hpctoolkit/-/merge_requests/570)

PR #574, opened 2022-05-10
Rework how we handle xed to link xed directly into libhpcrun.so and

mwkrentel:
libhpcrun.o.  This allows us to hide the xed symbols inside our
library in case the application also uses xed.

This now requires xed to be built statically (.a) with fPIC.

----------

@jmellorcrummey  This is the fix for the xed symbol clash.
Before merging into something, I want to test three things:

1. test on ppc and arm (me)
2. test on theta with static hpcprof-mpi (me)
3. verify that it actually does fix the xed clash with 2022.04.17 (johnmc)


jmellorcrummey (approved):
LGTM.

PR #577, opened 2022-05-13
Tune the static TLS block size when using LD_AUDIT

blue42u:
Glibc has a bug where the static TLS block is not allocated large enough when
LD_AUDIT is specified, causing an uncatchable crash before the application starts.
This commit temporarily avoids the issue by setting GLIBC_TUNABLES to "tune"
the "surplus" added to the block to 64KB (from 1664B).
An --auditor-static-tls-size option is added to hpcrun to control this size.

As a side effect of this, the memory usage per thread increases by 63KB (by default),
however as the additional pages are not touched there is little performance hit on x86-64.

***

cc @mwkrentel 

I'd like to backport this change to `master` after merge, to be included in the potential upcoming patch release.

### DIFF
--- a/doc/manual/hpcrun.tex
+++ b/doc/manual/hpcrun.tex
@@ -174,6 +174,14 @@ The following options direct \hpcrun{} to adjust the strategy it uses for monito
                        it is believed that the rewriting is causing the
                        application to fail.
                        
+\item{{\tt --auditor-static-tls-size <size>}}
+                       By default, \hpcrun{} uses \verb|LD_AUDIT| to track dynamic library
+                       operations. When using an auditor, a bug in glibc causes
+                       the size of the static TLS block to be severely underestimated.
+                       By default, \hpcrun{} will add 64KB to the static TLS to
+                       avoid this. This option can be used to adjust this amount,
+                       or can be \verb|none| to disable the tuning completely.
+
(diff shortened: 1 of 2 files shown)


jmellorcrummey (commented):
The patch looks complete. I read the documentation, help messages, and scripting code and all looks good.

I have not tested this, but I assume the GLIBC_TUNABLES works everywhere, meaning RHEL, Suse, Debian, ...

Once you confirm that either (a) you are certain of this because you've seen the way the GLIBC code does this, or (2) that you've tested this on various platforms and confirmed the effect, then I am fine with the patch.

blue42u:
The patch to add the tunable was added as a side-effect of fixing [BZ#25051](https://sourceware.org/bugzilla/show_bug.cgi?id=25051), which happened in Glibc 2.32. It was then backported by some of the distros, thus my confusion. Specifically:
 - Debian: backported in 11 (Bullseye)
 - Ubuntu: new enough Glibc in 20.10 (Groovy)
 - Red Hat: backported in 8.4
 - OpenSUSE: new enough Glibc in tumbleweed (rolling)

In conclusion, this patch will help none of the supercomputers I can access right now. Sigh. I'll close this PR for now, let's plan to fix this with the next auditor rewrite.

PR #578, opened 2022-05-18
Unwind callers not returns

### LINKED ISSUE #540: Fix caller Contexts off-by-one
Run records call stack unwinds in the CCT, which on (some? all?) architectures refers to the sequence of return addresses rather than the sequence of call instructions. Prof2 when parsing the CCT treats them as call addresses, causing weird calling contexts to appear in long (and often C++) callchains.

This will require an investigation into how unwinds turn into CCTs in Run.

Discussed with @jmellorcrummey

blue42u:
When unwinding a stack frame, the PC in the resulting frame is a return address, rather
than a call address. If the compiler did not expect the call to return, the compiler may
pack instructions in a way such that the unwind recipe for the return address is not the
same as the call address.

This commit changes the semantic of hpcrun_unw_step to always result in a "caller" frame,
with an approximate call address of (return address - 1), instead of the "return" frame
returned previously. This simplifies the incomplete logic added in b99930f9b3d4bb3d9e384e9f91e212cc0df53d53
and extends it to apply to the generic libunwind and Power cases as well.

The x86 native unwinder is unchanged in this commit, it fails to unwind even in the
microtest case used to validate this patch.

Fixes #540

### DIFF
--- a/src/tool/hpcrun/lush/lush.c
+++ b/src/tool/hpcrun/lush/lush.c
@@ -411,8 +411,6 @@ lush_step_pchord(lush_cursor_t* cursor)
 lush_step_t
 lush_forcestep_pnote(lush_cursor_t* cursor)
 {
-  int steps_taken = 1;
-
   if (lush_cursor_is_flag(cursor, LUSH_CURSOR_FLAGS_END_PPROJ)) {
     return LUSH_STEP_END_PROJ;
   }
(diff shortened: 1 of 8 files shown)


blue42u:
@jmellorcrummey Altered to only decrement PC when PC is non-0.

jmellorcrummey (approved):
LGTM

PR #588, opened 2022-06-08
hpcstruct: no <C> tags for CPU binaries

jmellorcrummey:
suppress generation of call site tags for CPU binaries.

First, nothing in hpctoolkit uses them so it is pointless to generate
them.

Second, hpcprof rejects any structure file with a <C> tag that
represents a call to an address not identified as a function in the
structure file. Some nested functions are not reported as functions,
so some structure files with <C> tags are rejected by hpcprof in
practice.


blue42u (approved):
LGTM

PR #590, opened 2022-06-09
refine messages from hpcprof and hpcstruct

jmellorcrummey:
hpcprof's default output directory is now an INFO message instead of a warning.
    
hpcstruct's NOTE messages are now INFO messages


blue42u (approved):
Adjusted so the `INFO:` messages during argument parsing work without also passing `-v`, and tweaked the message a bit to be more concise.

PR #592, opened 2022-06-11
hpcrun: add unwinding support for Intel CET

jmellorcrummey:
Add unwinding support for Intel Control-flow Enforcement Technology (CET), which may mark a branch target with the 'endbr64' instruction.

With Intel's compilers, we sometimes encounter routines marked with endbr64 that have symbols at an address 4 bytes less than where DWARF FDEs begin for the function. This disrupts their lookup with libunwind. For that reason, we consider adjusting the starting address of a function. If we fail to build unwind intervals for a routine from FDEs, then if
the routine begins with an 'endbr64' instruction, adjust the start of the routine to after the instruction and try building unwind intervals again.


jmellorcrummey:
Tested with Intel's ifort and icpx compilers and found to be effective.

PR #597, opened 2022-06-13
correct opencl headers path spec in configure.ac

jmellorcrummey:


mwkrentel:
Now that you changed configure to add `include`, that leaves the help
string inconsistent.  Let's change that to something like:

```
AC_ARG_WITH([opencl],
  [AS_HELP_STRING([--with-opencl=PATH],
      [path to opencl install prefix, only uses include dir for headers])],
```

jmellorcrummey:
messages updated to reflect what the --with-opencl path means.

PR #598, opened 2022-06-14
Add a configure check for C++17

blue42u:
Reuses and replaces the previous C++11 check. There are two main differences from the previous logic:
 - GNU dialects (`-std=gnu++17`) are never checked. I don't think we should ever need GNU extensions for C++, so this check enforces that.
 - ~~The "no additional flag" case is never checked. g++ enables GNU extensions by default, so this makes sure we always get the ISO dialect. Setting `CXX17_FLAG=' '` can be used to override this.~~ 
   The "no additional flag" case is not checked unless CXX17_FLAG=none, and even then is checked last. g++ enables GNU extensions by default, so this makes sure we always get the ISO dialect unless otherwise needed or specified.

Resolves the complaint about hardcoding flags in https://github.com/HPCToolkit/hpctoolkit/pull/570#issuecomment-1151353816. Compiles with GCC 7-12.


mwkrentel:
@blue42u  If it were me, I'd prefer to keep 'none' for two reasons.

1. Depending on the version of compiler, it may support the desired
features without any flag but not recognize `-std=c++17`, especially
for non-gnu compilers.  Yes, I know we don't really support non-gnu,
but I try not to put extra roadblocks in the way.

2. Depending on how someone builds hpctoolkit, there may already be a
valid flag in CXXFLAGS and adding an extra flag may interfere or
change the semantics.  This was the reason for using 'none' before any
flags for the c++11 case.  If someone already wrote a flag and it
works, then don't undo it.

Anyway, if it comes down to add no flag or fail the build, I'd rather
not fail the build.

But I suppose we could at least use 'none' but move it to the end.

----------

I don't think this will do quite what you want.

```
  AC_MSG_WARN([unable to find CXXFLAGS for std C++17])
  AC_MSG_ERROR([set the CXX17_FLAG variable and rerun configure])
  AC_MSG_ERROR([if no flag is needed, set CXX17_FLAG=' ' and rerun configure])
```

`AC_MSG_ERROR` causes configure to die, so the second one will never
be printed.  That's the reason for making the earlier ones `WARN`.

Of course, if I can convince you to add none, then the second `ERROR`
is unnecessary.

mwkrentel (changes requested):
see other comment

blue42u:
@mwkrentel 
IMHO if someone is using a compiler that doesn't understand GNU-style arguments, they should really tell us what we need to flip to get what we want (or use a wrapper script to convert the arguments, e.g. [Kokko's `nvcc_wrapper`](https://github.com/kokkos/kokkos/blob/master/bin/nvcc_wrapper)). If we don't need to flip anything the user should be explicit about telling us (`CXX17_FLAG=' '` is basically "trust me, I know what I'm doing"). Failing in `./configure` is (AFAIK) conventionally the spot for these sorts of "hey, I'm not sure what you're doing, please explain" situations.

On the other hand, most of our users will be compiling with Spack. I argue they will have far worse issues trying to get a non-GNU compiler through our dependencies (e.g. Elfutils), but editing Spack recipes is admittedly not fun. Maybe our Spack recipe could set a `CXX17_ALLOW_NO_FLAG=1` argvar when the `cxxflags` variant is non-empty, and `none` would be checked last only if that argvar is set?

* * *
Ack. This is my lack of real Autotools knowledge showing through. Will fix once the other comment is resolved.

mwkrentel:
@jmellorcrummey @blue42u 
I still don't see why leaving 'none' as the last test could cause any
problems.  So I'll repeat my argument and let @jmellorcrummey be the jury.

The only case where this matters is: suppose adding `-std=c++17` fails
the test, but leaving CFLAGS unchanged makes it succeed.  That's
unlikely, but if it happens, then IMO there's no point in failing the
build.  The test passed, so continue.

Note: the real point of the 'none' case is to allow the proper
flag to be passed in via CFLAGS.  If the test passes with the given
CFLAGS, then use it.  That's how a spack build would work.

Ok?

blue42u:
> The only case where this matters is: suppose adding -std=c++17 fails
the test, but leaving CFLAGS unchanged makes it succeed. That's
unlikely, but if it happens, then IMO there's no point in failing the
build.

The only reason why it would be useful to fail the build is if you're developing. We (attempt to) have rules that our code is compatible with (almost) C++17 (for GCC 7.x) and is ISO-compliant (to ease using non-GNU compilers). Passing `-std=c++17` enforces (most of) these constraints, if 'none' is an option `./configure` could pass without this enforcement (i.e. the test was faulty and failed for unintentional reasons).

It's roughly the same rationale as having `-Werror` on by default: it's far cheaper than waiting for someone else (or CI) to report a compile failure (or worse).

Obviously compiling for Spack is a bit different, there the build must pass so 'none' needs to be an option. Setting `CXX17_FLAG=' '` basically does this, except it's at the beginning (and I would prefer it at the end).

Anyway, that's my opinion, I'll adjust the patch according to what @mwkrentel and/or @jmellorcrummey wants.

jmellorcrummey:
@mwkrentel I don't understand how the 'none' was helpful before and why having it eliminated is a problem. Perhaps the three of us should chat for a few minutes on Monday so we can do something with this PR.

mwkrentel:
@jmellorcrummey `none` was for the case that the appropriate flag was
already passed in via CXXFLAGS.  `none` doesn't mean "use no flags",
it means, "try the existing flags unchanged."  This is how spack would
pass flags into autotools.

If `none` is the last thing to try, it's unlikely that it will ever be used.
But it does guarantee that if CXXFLAGS provides a valid flag, then
the configure test will succeed.

blue42u:
@jmellorcrummey @mwkrentel What's left for this PR? I'd like to get it merged to unblock https://github.com/HPCToolkit/hpctoolkit/pull/570.

jmellorcrummey:
@blue42u Before completely changing hpcstruct, we are currently working to address failures at LLNL. 

- I have new patches for Dyninst that fix data races when analyzing NVIDIA binaries in parallel: https://github.com/dyninst/dyninst/pull/1284
- Marty committed a patch to hpctoolkit master to make sure that corrupted entries in the structure cache are eliminated https://github.com/HPCToolkit/hpctoolkit/commit/5f8eafff715bcf8332afb90d42fb2a1a40c191a4

Once I am done with my tri-labs tutorial this week, I can work on reviewing and merging this commit.

blue42u:
Updated, now setting `CXX17_FLAG=none` tries the "no flag" case after all hardcoded cases have been attempted. (Of course `CXX17_FLAG=' '` still functions to test the "no flag" case first.)

Also rebased onto the latest `develop`.

mwkrentel:
No, this is still not what I had in mind.  The CXX11 or CXX17 flag was
just a safety valve in case that some compiler used an unusual flag
not in the list.  (Btw, this was before spack and spack would never
use it.)

It shouldn't be necessary to set CXX11_FLAG to test the 'none' case.

What I had in mind was 'none' was in the search list (at the end, if
you wish) and this case would trigger the empty string as the added
flags.  The point of this case is if some newer compiler automatically
includes C++17 support without any additional flags.

See how master does this.

blue42u:
> It shouldn't be necessary to set CXX11_FLAG to test the 'none' case.

> The point of this case is if some newer compiler automatically
includes C++17 support without any additional flags.

> See how master does this.

I'm aware of how `master` (and `develop`) does this, there's a note in the OP describing the change in logic that this PR implements. I don't agree with the original behavior in `master`. If you don't mind me spouting off for a few minutes:

IMO I don't want to include a "none" case by default. IMHO the default settings in `./configure` should be tuned for developers instead of for end users, and I believe this is appropriate: AFAIK in recent years (i.e. ever since Spack) very few people have *manually* built HPCToolkit from source outside of developers doing development. (Even I use Spack when in a non-development setting.) The requirement for our newer code is that it conforms to ISO C++17, thus `./configure` should by default not pass without having a `-std=` argument to enforce this requirement. Especially since late-model g++ defaults to `-std=gnu++17` which allows potentially unportable GNU extensions.

This makes it far easier for us to write cleaner code and finds issues in code far earlier. I don't want anyone else to spend hours finalizing a PR only to have the CI (or worse, someone in the group) root out what amounts to a syntax quirk not understood by GCC 7.x. (...Never again, please.)

Spack needs to build HPCToolkit for the end user, so it must still compile in the case where no `-std=` argument is needed (non-GNU compiler, set in `CXXFLAGS`, or otherwise). `CXX17_FLAG=none` provides this functionality in a way I prefer, as a fallback if no known `-std=` arguments seem to work. `CXX17_FLAG=' '` provides this functionality closer to how `master`/`develop` do, where `-std=` flags are only attempted if C++17 support doesn't appear to be present otherwise. Either of these could be added to the Spack recipe for `@develop` for eventual inclusion in a full release with almost zero disturbance.

**TL;DR:** I think we're disagreeing on the primary purpose of `./configure`('s default behavior). I believe it should - and is appropriate to - be for developers first and for end users second.

@mwkrentel If you can't agree with my change to the behavior, I'll circle back tonight and adjust this patch to be closer to the original. But again, I - and once CI is enabled, I believe the entire group - will be happier campers with this change.

blue42u:
[This repository has moved to GitLab, please continue this PR there.](https://gitlab.com/hpctoolkit/hpctoolkit/-/issues/598)

blue42u:
The previous link was incorrect, apologies. The corrected link is [here.](https://gitlab.com/hpctoolkit/hpctoolkit/-/merge_requests/598)

PR #603, opened 2022-06-20
Update args.cpp

Jokeren:
There might be no "measurements" in measurement dir's name.


blue42u (approved):
Unable to test, but LGTM.

PR #605, opened 2022-06-25
hpcstruct: set OMP_STACKSIZE

jmellorcrummey:
ensure that OpenMP workers have the same size stack as the main
thread by setting their stack size to "ulimit -s" kilobytes

### DIFF
--- a/src/tool/hpcstruct/hpcstruct.in
+++ b/src/tool/hpcstruct/hpcstruct.in
@@ -44,5 +44,13 @@ case "$libexec_dir" in
     * )  libexec_dir="${HPCTOOLKIT}/${libexec_dir}" ;;
 esac
 
+#------------------------------------------------------------
+# Ensure that the stack for each worker thread is the same
+# size as the stack of the master thread. This will avoid
+# having a multi-threaded execution run out of stack when a
+# single-threaded execution would not.
+#------------------------------------------------------------
+export OMP_STACKSIZE=`ulimit -s`k
+
 export LD_LIBRARY_PATH="${ext_libs_dir}:${libcxx_path}:${LD_LIBRARY_PATH}"
 exec "${libexec_dir}/${binary_name}" "$@"
--- a/src/tool/hpcstruct/hpcstruct.in
+++ b/src/tool/hpcstruct/hpcstruct.in
@@ -83,10 +83,10 @@ do
 done
 
 #------------------------------------------------------------
-# Ensure that the stack for each worker thread is the same
-# size as the stack of the master thread. This will avoid
-# having a multi-threaded execution run out of stack when a
-# single-threaded execution would not.
+# 3 strategies to set OMP_STACKSIZE, in order of priority
+# 1. -s argument
+# 2. use existing value of OMP_STACKSIZE, if any
+# 3. default OMP_STACKSIZE to 8M (8MB)
 #------------------------------------------------------------
 if test "x$stacksize" != x; then
   export OMP_STACKSIZE=$stacksize
--- a/src/tool/hpcstruct/hpcstruct.in
+++ b/src/tool/hpcstruct/hpcstruct.in
@@ -44,13 +44,54 @@ case "$libexec_dir" in
     * )  libexec_dir="${HPCTOOLKIT}/${libexec_dir}" ;;
 esac
 
+die()
+{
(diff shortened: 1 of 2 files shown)


jmellorcrummey:
I wrote a preloaded wrapper for pthread_create to check the stack size for GNU OpenMP. Without setting OMP_STACKSIZE, pthread_getstacksize reports that no stack size is specified. 

According to the man page for pthread_create, the default size of pthread stacks is architecture dependent and is 4MB on Power and 2MB on x86_64. The man page doesn't specify for aarch64.

This PR sets the default size for a pthread to be the stack size set for the main thread. This should avoid surprises. Yesterday, we had a report from Matt at LLNL that hpcstruct worked with one thread on a large binary but failed in multithreaded mode. A bit of investigation showed an OpenMP worker thread running out of stack. This patch should make each OpenMP worker just as capable as the main thread. 

Using my pthread_create wrapper that inspects pthread attributes, I can see that this wrapper results in a stack size attribute setting the desired size. The command 'ulimit -s' works for sh, the shell running the hpcstruct script.

mwkrentel:
> This PR sets the default size for a pthread to be the stack size set for the main thread.

@jmellorcrummey `ulimit -s` doesn't necessarily return the main stack size.  
The hard limit is normally unlimited, so I as a user can set the soft limit to
anything, including way too high, way too low, or 'unlimited.'

I was going to suggest testing the answer from ulimit to make sure
that it's a number.

```
stksize=`ulimit -s`
if expr 1 + "$stksize" >/dev/null 2>&1 ; then : ; else
    stksize=8192
fi
export OMP_STACKSIZE="$stksize"
```

But as I said, ulimit could be way off.  And I wouldn't normally
think of controlling the openmp stack size via ulimit.

I would also suggest using `OMP_STACKSIZE` if it's already set.
Maybe just use OMP_STACKSIZE and then 8192 if that's unset.

jmellorcrummey:
How about a different strategy:
- add a -s argument to hpcstruct that allows one to specify the stack size of the OpenMP threads in MB
- use the MB specified by a -s argument, if any
- use 8MB if unspecified

mwkrentel:
Ok, except that limits like this are more commonly written in K
than M, unless you want to allow a units value (K, M, G) and parse it.

jmellorcrummey:
New implementation:

in order of priority
- consult '-s' option with which user can specify size
- use value of OMP_STACKSIZE (if present)
- set OMP_STACKSIZE=8M

Note: hpcstruct now processes the '-s' option in the shell
script and the rest of the options in the program. this is
an unholy mix. however, setting OMP_STACKSIZE in the
application with putenv doesn't work. Perhaps OMP_STACKSIZE
is inspected by a library constructor. this will do for now
as a hot fix.

mwkrentel:
A few things, but basically I approve.

(1) As you noticed, the default unit for OMP_STACKSIZE is K.

(2) According to this page for Openmp, there is no internal function
like `omp_set_num_threads()` to set the thread stack size.  Probably
because Openmp wants the value early.

https://www.openmp.org/spec-html/5.0/openmpsu32.html#x51-710002.5.3

I'd bet you could still do it inside the hpcstruct binary, but you'd
have to start hpcstruct, do putenv() and then re-exec() the binary.
All being very careful not to do anything significant before the
exec().  Basically, that's what the shell script is doing.

So, if you absolutely, positively want to move this into the binary
and eliminate the shell script, it's possible, but a bit awkward.
Or, maybe this is a standard thing to do.

(3) The way you parse the command line for `-s num` is vulnerable to
args with spaces in them.  This is a persistent problem for shell
scripts.  But I can show you how to be a bit more robust.

Basically, you create a marker, some string that would never be on the
command line, add it to the end of the command line (with set),
iterate through the args, but instead of `args="$args $arg"`, you
re-add it with set:

```
set -- "$@" "$arg"
```

This is a common shell trick.  Or, you could use a language (perl,
python, even bash) that supports actual arrays.  (Bourne shell has
only one array, namely $@).

Or you could leave the command line intact and teach hpcstruct to
parse and ignore `-s`.

But I'm Ok if you want to commit this now and clean it up later, that
is, assign it to me.  :-)

jmellorcrummey:
@mwkrentel I am happy for you to approve this later. Right now, I want this commit to fix the problem at LLNL.

PR #616, opened 2022-06-29
clean up --with-level0 in configure.ac

jmellorcrummey:
- check for the presence of a level_zero include file
- eliminate excess substitutions


mwkrentel:
(1) I realize this isn't part of the patch, but rather than 

```
if ( echo "${withval}" | grep -v "^/" >/dev/null 2>&1 ); then
```

to test for an absolute path, it's a lot simpler (and doesn't require
outside utilities) to use `case`.

(2) In this code section,

```
 else
   AC_MSG_ERROR([unable to find level_zero/ze_api.h in: ${OPT_LEVEL0}/include])
   OPT_LEVEL0=
 fi
```

`AC_MSG_ERROR` exits the process, so anything after that will never happen.

Otherwise, I approve.

jmellorcrummey:
Comments noted. I'll commit this PR, which affects every Makefile.in. The other things you mentioned are local to configure and could be done as part of later configure.ac tinkering.

PR #617, opened 2022-06-29
Pr fix perf

laksono:
Fix perf_event sample source when profiling app with fork
Adapt the solution from itimer.c to solve problems when profiling forked executables.
Caveats:
- This branch is not compilable with gcc 7.3.0, but works fine with 8.x or newer.
- This branch doesn't fix the problem with PAPI_* when profiling forked executables. My understanding is PAPI hides the signal handling inside PAPI code and doesn't expose it. It requires more time to investigate.


jmellorcrummey (approved):
LGTM (for now).

I wonder whether we can move the itimer and perf  monitor_real_pthread_sigmask calls to initialization and shutdown and finalization rather than incur the cost every time we start and stop the sample source? That would be more efficient, if safe.

PR #618, opened 2022-06-30
hpcrun: Fix PAPI issue with forks

laksono:
The reason why hpcrun "crashed" when profiling forked executables with
PAPI is because in the child process, when hpcrun calls
PAPI_thread_init, PAPI library is not initialized although it's
initialized in the parent process.

The solution is to call PAPI_library_init before calling
PAPI_thread_init when a thread is created.

### DIFF
--- a/src/tool/hpcrun/sample-sources/papi-c.c
+++ b/src/tool/hpcrun/sample-sources/papi-c.c
@@ -286,7 +286,18 @@ METHOD_FN(thread_init)
   TMSG(PAPI, "thread init");
   if (papi_unavail) { return; }
 
-  int retval = PAPI_thread_init(pthread_self);
+  // Need to call PAPI_library_init before calling PAPI_thread_init
+  // Somehow, when a process is forked, the child process doesn't
+  // inherit PAPI library init and thus we have to explicitly call it
+  // here. 
+  int retval = PAPI_library_init(PAPI_VER_CURRENT);
+  TMSG(PAPI_C,"PAPI_library_init = %d", retval);
+  if (retval < 0) {
+    EEMSG("PAPI_library_init NOT ok, retval = %d", retval);
+    monitor_real_abort();
+  }
+
+  retval = PAPI_thread_init(pthread_self);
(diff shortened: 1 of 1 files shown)


jmellorcrummey (commented):
Is the right solution to jam a call to PAPI_library_init into the thread_init method?

Or, would the right solution be to make sure that the init method is called before thread_init?

laksono:
> Is the right solution to jam a call to PAPI_library_init into the thread_init method?
> 
> Or, would the right solution be to make sure that the init method is called before thread_init?

I am still working on it. 
It turns out this solution works for loop-suite tests, but not for caffeine.

mwkrentel:
@laksono, do you have a small example that produces the problem here?
That is, fork() from a threaded program?
Btw, when it fails, how does it fail?

I'm going to have to plead ignorance.  Long, long ago, back in the
perfmon/perfctr days, I remember we worked with PAPI to get the
semantics of PAPI_overflow() straight across fork() and exec().  But
that was probably not with a threaded program.

I'm suspicisous of doing things like PAPI_init() and PAPI_shutdown()
in threaded programs because those will affect the entire process.

laksono:
> @laksono, do you have a small example that produces the problem here? That is, fork() from a threaded program? Btw, when it fails, how does it fail?

Unfortunately we don't have a small test program that mimics the problem with caffeine.

The problem is that it fails in the child process. 
Somehow the `eventset` variable defined in the parent is not "recognized" in the child process.
Hence when in the child process we call `PAPI_list_events`, it returns error that the `eventset` is not initialized.

mwkrentel:
@laksono Do you know more about the sequence inside caffeine and where
it goes wrong?  I'm wondering if monitor-debug or the right dd
variable would show us exactly the sequence that happens in the child.

And do you know that caffeine is threaded and spawns threads before
the fork()?

Anyway, rather than just always re-do the library init, I'd feel more
comfortable with an atomic variable that tells us if library init has
happened.

Maybe library init is run but turns stale, or maybe the threads are
started in a context where library init hasn't happened that path.

PR #622, opened 2022-07-05
Support call traces

### LINKED ISSUE #619: prof2 shows call path within GPU kernels in trace when it shouldn't
When measuring and analyzing LAMMPS with prof2 in the johnmc branch, I found that prof2 is adding detail about call paths internal to the GPU to the heterogeneous call chain for a kernel call in the GPU trace. We don't trace information within GPU kernels so we can't display it. 

The figure below illustrates the problem. Notice that the trace shows a deep call path WITHIN the GPU kernel so that all of the elements in the trace appear to be in `<unknown procedure>` (green at the bottom of the call chain) rather than in `Kokkos::impl::cuda_parallel_launch_constant_memory<...>() `(the item on the path immediately below `<gpu kernel>`. Even when PC sampling, we don't trace internal details of kernels because we get GPU PC samples as a histogram (with the time dimension collapsed out).

image

The database that yielded this figure can be found in ufront.cs.rice.edu:~johnmc/bugs/prof2-trace-bug.tgz

blue42u:
CPU traces are composed of a series of "samples" taken from every running thread. In contrast, GPU traces are (currently) composed only of top-level operations such as `<gpu copy>` or kernel launches. This PR better supports the latter case by marking such "call traces" and preventing deep lexical expansion (of inlined call chains) in Prof2 for these traces.

Fixes #619.

### DIFF
--- a/src/lib/prof-lean/hpcrun-fmt.h
+++ b/src/lib/prof-lean/hpcrun-fmt.h
@@ -811,6 +811,7 @@ typedef struct hpctrace_hdr_flags_bitfield {
 // Substitute bit fields with macros
 #define HPCTRACE_HDR_FLAGS_DATA_CENTRIC_BIT_POS 0U
 #define HPCTRACE_HDR_FLAGS_LCA_RECORDED_BIT_POS 1U
+#define HPCTRACE_HDR_FLAGS_CALL_TRACE_BIT_POS 2U
 
 #define HPCTRACE_HDR_FLAGS_GET_BIT(flag, pos) \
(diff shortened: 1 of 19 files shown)


jmellorcrummey (approved):
LGTM. It is awkward when we need to replace a boolean with a 3+ valued type. If only we envisioned that earlier, a lot of editing could have been avoided :-).

I don't understand the changes to hpctracedb2.cpp, but then I don't understand the way that the data is being processed at all. I'll trust you on this.

PR #623, opened 2022-07-14
Handle dlopen in threads without thread_data

wyphan:
This was observed when trying to profile PeleC with AMD GPU on ufront.cs.

### DIFF
--- a/src/tool/hpcrun/safe-sampling.h
+++ b/src/tool/hpcrun/safe-sampling.h
@@ -167,6 +167,10 @@ hpcrun_safe_enter_async(void *pc)
 static inline void
 hpcrun_safe_exit(void)
 {
+  if (! hpcrun_is_initialized() || ! hpcrun_td_avail()) {
+    return 0;
+  }
+
   thread_data_t *td = hpcrun_get_thread_data();
 
   td->inside_hpcrun = 0;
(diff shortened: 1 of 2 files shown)


src/tool/hpcrun/safe-sampling.h
```diff
@@ -167,6 +167,10 @@ hpcrun_safe_enter_async(void *pc)
 static inline void
 hpcrun_safe_exit(void)
 {
+  if (! hpcrun_is_initialized() || ! hpcrun_td_avail()) {
+    return 0;
```
Jokeren, line 171: Since the return type of this function is void. I'll suggest you just use `return;`

src/tool/hpcrun/unwind/common/uw_recipe_map.c
```diff
@@ -588,15 +588,18 @@ uw_recipe_map_notify_unmap(load_module_t* lm)
...
+
+    // we need to confirm that the hashtable is allocated. if we are
+    // exiting very early (e.g. when a metric is not recognized), which
+    // means that hpcrun_init_internal has not yet been called, the hash
+    // table has not yet been allocated.
+    if (td->uw_hash_table) uw_hash_delete_range(td->uw_hash_table, start, end);
+  }
 
   uw_recipe_map_report_and_dump("*** unmap: after poisoning", start, end);
+
```
Jokeren, line 602: Remove empty line

src/tool/hpcrun/safe-sampling.h
wyphan: Whoops, my bad. I guess that's called copy-pasting. 🙈

jmellorcrummey (approved):
LGTM.

Thanks for the careful review @Jokeren!

PR #625, opened 2022-07-17
hpcstruct: don't SEGV when analyzing an unreadable file

jmellorcrummey:
computing a hash string of an unreadable file would return NULL rather than a string of hex digits. change so it computes "", which avoids a the SEGV.

### DIFF
--- a/src/tool/hpcstruct/SingleBin.cpp
+++ b/src/tool/hpcstruct/SingleBin.cpp
@@ -158,6 +158,15 @@ doSingleBinary
   string cache_flat_entry;
   string cache_directory;
 
+  // Make sure the file is readable
+  if ( access(binary_abspath.c_str(), R_OK) != 0 ) {
+    cerr << "ERROR -- input file " << args.in_filenm.c_str() << " is not readable" << endl;
+    if ( args.is_from_makefile == true ) {
+      cerr << "CACHESTAT (Input file is not readable) " << endl;
+    }
+    exit(1);
+  }
+
   if (args.nocache) {
     // the user intentionally turned off the cache
     args.cache_stat = CACHE_DISABLED;
--- a/src/tool/hpcstruct/Structure-Cache.cpp
+++ b/src/tool/hpcstruct/Structure-Cache.cpp
@@ -448,6 +448,10 @@ hpcstruct_cache_hash
 )
 {
   char *eh  = elf_hash(binary_abspath);
+
+  // handle NULL
+  if (eh == 0) return strdup("");
+
   return eh;
 }
 


martyitz:
You asked me to investigate, and I did.  Why did you investigate and propose a fix?

I have what I think is a better fix.  It is not in a branch, but in my copy of
master.  The change is in SingleBin.cc which checks at the top of the file for access to the target.

What does your code report for a measurements directory?  Mine reports
    43  begin parallel analysis of CPU binary postrept (size = 17336, threads = 2)
     44    end parallel analysis of CPU binary postrept (Input file is not readable)

For a single binary, it says:
     ERROR -- input file /home/msi3/gits/minitest/bin/postrept is not readable

My change is:
diff --git a/src/tool/hpcstruct/SingleBin.cpp b/src/tool/hpcstruct/SingleBin.cpp
index e6be3e6d2..778eef6d8 100644
--- a/src/tool/hpcstruct/SingleBin.cpp
+++ b/src/tool/hpcstruct/SingleBin.cpp
@@ -158,6 +158,15 @@ doSingleBinary
   string cache_flat_entry;
   string cache_directory;
 
+  // Make sure the file is readable
+  if ( access(binary_abspath.c_str(), R_OK) != 0 ) {
+    cerr << "ERROR -- input file " << args.in_filenm.c_str() << " is not readable" << endl;
+    if ( args.is_from_makefile == true ) {
+      cerr << "CACHESTAT (Input file is not readable) " << endl;
+    }
+    exit(1);
+  }
+
   if (args.nocache) {
     // the user intentionally turned off the cache
     args.cache_stat = CACHE_DISABLED;

jmellorcrummey:
this is not the issue that i asked you to investigate. 

the issue i asked you to investigate was to see if we could replicate why hpcrun failed only if
(1) monitoring amd GPU activity in ARES, and (2) spawning an unreadable binary. i asked you to see if we could reproduce this. we might have to have cray mpi as part of the mix. i think that cray mpi may scrub open file descriptors on fork and that may be the cause of the error with rocm.

this fix to hpcstruct is a related issue, if you try to analyze an unreadable binary. certainly, the result shouldn’t be segv. should it be an error at all? that may be awkward with an unreadable library (is such a thing usable?) as one of many libraries in an application. if analysis of an unreadable binary triggers an error rather than a warning, this could disrupt a parallel hpcstruct of an executable and a set of libraries.

martyitz:
I told you about this error and that I was investigating, and I asked you if this was the problem
that MartyM reported.  You did not reply.

> On Jul 16, 2022, at 7:22 PM, John Mellor-Crummey <[notifications@github.com](mailto:notifications@github.com)> wrote:
> 
> this is not the issue that i asked you to investigate.
> 
> 

> the issue i asked you to investigate was to see if we could replicate why hpcrun failed only if
> (1) monitoring amd GPU activity in ARES, and (2) spawning an unreadable binary. i asked you to see if we could reproduce this. we might have to have cray mpi as part of the mix. i think that cray mpi may scrub open file descriptors on fork and that may be the cause of the error with rocm.
> 

 
That’s the answer I was waiting for.  When I created the infrastructure to investigate the issue, as we had agreed
I should do, I stumbled upon the SEGV, and devised a fix for it.

Tomorrow, I will, at the risk of crashing ufront, investigate the hpcrun failure with AMD GPU usage, and a call to system for an unreadable binary.

> this fix to hpcstruct is a related issue, if you try to analyze an unreadable binary. certainly, the result shouldn’t be segv. should it be an error at all? that may be awkward with an unreadable library (is such a thing usable?) as one of many libraries in an application. if analysis of an unreadable binary triggers an error rather than a warning, this could disrupt a parallel hpcstruct of an executable and a set of libraries.

My hpcstruct only prints ERROR when invoked on a single binary.
The log file also shows that hpcprof prints two warnings about the unreadable binary:

    WARNING: Unable to open input file: /home/msi3/gits/minitest/bin/postrept (Permission denied)
    WARNING: Cannot fully process samples for load module /home/msi3/gits/minitest/bin/postrept: binary file not found
 
The first is correct; the second is not.  The file is found, but unreadable.

martyitz:
I committed and pushed my change to [pr-hpcstruct-unreadable](https://github.com/HPCToolkit/hpctoolkit/tree/pr-hpcstruct-unreadable).  As far as I can tell my copy has both our changes, which is like having a belt and
suspenders.  I assigned the bug to you -- commit with either or both fixes.

martyitz (approved):

martyitz:
I told you about this error and that I was investigating, and I asked you if this was the problem
that MartyM reported.  You did not reply.

> On Jul 16, 2022, at 7:22 PM, John Mellor-Crummey ***@***.***> wrote:
> 
> this is not the issue that i asked you to investigate.
> 
> the issue i asked you to investigate was to see if we could replicate why hpcrun failed only if
> (1) monitoring amd GPU activity in ARES, and (2) spawning an unreadable binary. i asked you to see if we could reproduce this. we might have to have cray mpi as part of the mix. i think that cray mpi may scrub open file descriptors on fork and that may be the cause of the error with rocm.
> 
 
That’s the answer I was waiting for.  When I created the infrastructure to investigate the issue, as we had agreed
I should do, I stumbled upon the SEGV, and devised a fix for it.

Tomorrow, I will, at the risk of crashing ufront, investigate the hpcrun failure with AMD GPU usage, and a call to system for an
unreadable binary.
> this fix to hpcstruct is a related issue, if you try to analyze an unreadable binary. certainly, the result shouldn’t be segv. should it be an error at all? that may be awkward with an unreadable library (is such a thing usable?) as one of many libraries in an application. if analysis of an unreadable binary triggers an error rather than a warning, this could disrupt a parallel hpcstruct of an executable and a set of libraries.
> 
 
hpcstruct only prints ERROR when invoked on a single binary.
The log file also shows that hpcprof prints two warnings about the unreadable binary:

    WARNING: Unable to open input file: /home/msi3/gits/minitest/bin/postrept (Permission denied)
    WARNING: Cannot fully process samples for load module /home/msi3/gits/minitest/bin/postrept: binary file not found
 
The first is correct; the second is not.  The file is found, but unreadable.

	Marty

PR #626, opened 2022-07-21
fix enumeration and display of AMD GPU metrics

jmellorcrummey:
AMD moved the location of rocprofiler's metrics.xml file.
use configure to find it. specify the file path in a new
autoconf substitution variable ROCM_PROFILER_METRICS.

fix the display of rocprofiler metric descriptions by
using display_event_info to wrap description text as
needed.

remove display of gpu=openmp metric option. it is enabled
by default.

note: while making necessary adjustments in hpcrun.in, clean up the formatting, which was a mess.


jmellorcrummey:
tested with rocm-5.1.0, rocm-5.2.0, and with --with-rocm-profiler path from 'spack find -p rocprofiler-dev'

review only the configure.ac file, the C files, and hpcrun.in for changes. configure and Makefile.in are all derived and don't need inspection.

mwkrentel (approved):
Ok, LGTM, I approve.

PR #628, opened 2022-07-28
roctracer: enable ACTIVITY_DOMAIN_HIP_OPS

laksono:
Tested with rocm/5.2.0:
Need to enable ACTIVITY_DOMAIN_HIP_OPS to trace gpus
if this is disabled, no profiles in gpu kernels


jmellorcrummey (approved):
Written and tested with Laksono

PR #629, opened 2022-07-29
improve roctracer trace lines: use device+queue

jmellorcrummey:
ensure devices have separate trace lines by propagating roctracer device_id and queue_id into hpctoolkit's context_id and stream_id.


jmellorcrummey:
co-developed with laksono.

PR #630, opened 2022-08-01
Fix the teardown of roctrace infrastructure to match setup in reverse…

martyitz:
Teardown is in the reverse order from the setup.


jmellorcrummey (approved):
LGTM.

PR #631, opened 2022-08-11
Rework the start-up sequence for the sample sources.  This fixes the

mwkrentel:
fork + papi problem where the PAPI source would fail and/or segfault
in a process that fork()s.  The key points are:

  1. If a process forks, stop the samples only in the thread that
  calls fork, and restart them after the fork.  Don't shutdown the
  source in the parent since that would kill samples in all threads.

  2. In the child after a fork, shutdown the sources and restart them
  from scratch.  The child has one thread with a new pid, so this is
  the safest way to start the samples.

  3. Treat every process as threaded and always call the thread-init
  method after library-init.  This doesn't cause any trouble and there
  are too many code paths to call thread-init selectively.

Note: we call library-init on all sources, whether used or not.  (PAPI
requires this.)  But we call thread-init only on the sources that are
used.  So, the sequence must be: library-init, process-event-list (to
determine which sources are used), and then thread-init.

Cleanup a couple of -dd messages.  Collapse PRE_FORK and POST_FORK
into just FORK.  There are very few of these anyway.

Replace PAPI_C with PAPI.  Most of the -dd messages in papi-c.c are
still PAPI.  There were only 3 uses of PAPI_C, all at init time.

----------

@jmellorcrummey @blue42u @Jokeren @wyphan 

I've tested this with PAPI, POSIX timers (realtime, cputime), CYLES.
It works fine with the non-gpu sources.

But this needs serious testing with the various GPU sources,
especially ones that spawn helper threads of their own.

Fundamentally, this is the right way to launch sample sources and to
dance around fork().  I'm just hoping that the GPUs are compatible
with it.


jmellorcrummey (commented):
I have significant concerns with this patch. The PR makes no remark about the fact that hpcrun_prepare_measurement_subsystem is now called in a non-child process as part of monitor_init_process. I think that will be a problem for AMD GPUs if (as I suspect) monitor_init_process is called when pthread_create is intercepted while creating a ROCm helper thread.

This should be tested with Marty's minitests. Marty should turn off the problematic AMD tests that can crash the kernel on ufront ...

jmellorcrummey:
@martyitz can you check out and test Mark's branch?

mwkrentel:
Sorry, I misunderstood what hpcrun_prepare_measurement_subsystem() is
used for.  The way all the various pieces interleave is a mess, and
it's not like there are any comments to identify the use case.

Tell me more about the init process, what that function is used for
and give me an example where it fails and I'll take another look.
Preferably something that doesn't crash the machine.

jmellorcrummey:
hpcrun_prepare_measurement_subsystem is used to turn on the measurement subsystem at a clean point. AMD creates a thread during hsa_init. When hsa_init launches ROCm's  thread for shepherding asynchronous operations, that is definitely not a clean point. We simply can't initialize AMD GPU monitoring in the middle of hsa_init. For that reason, the prior code in master delayed initalization of hpctoolkit's measurement system with hpcrun_prepare_measurement_subsystem until main is called. In a child that is forked but not execed, that point may not be reached again, so hpcrun_prepare_measurement_subsystem in the child was done in monitor_init_process.

On ufront, you can use /scratch/johnmc/HIP-examples/

There is a gpu-burn code built there that is suitable.

wyphan:
@mwkrentel Just tested with Caffeine at ufront.cs, looks like it's working!

wyphan:
With loop-suite tests 8.1 and 12, `hpcprof` emits the following warnings, but otherwise it works just fine.
```console
hpcprof -o loop.d loop.m
WARNING: Merging profiles with different names: `loop' and `child'!
WARNING: Merging profiles with different paths: `/home/wp11/work/hpctoolkit-tests/applications/loop-suite/12.loop-fork-exec-papi/loop' and `/home/wp11/work/hpctoolkit-tests/applications/loop-suite/12.loop-fork-exec-papi/child'!
WARNING: Merging profiles with different names: `loop' and `child'!
WARNING: Merging profiles with different paths: `/home/wp11/work/hpctoolkit-tests/applications/loop-suite/12.loop-fork-exec-papi/loop' and `/home/wp11/work/hpctoolkit-tests/applications/loop-suite/12.loop-fork-exec-papi/child'!
```

martyitz:
If you specify a -o argument to hpcrun, and the target process creates descendants, they will be written to
the -o directory, no matter what the executables were named.  One of my minitest runs creates
run files for the target, sngl.tf.cuda.gcc.nv, and also for bash and postrept, invoked by the target.

If you don't specify a -o argument, separate measurement directories are created for the different executables.

mwkrentel:
@jmellorcrummey   Ok, I'm deducing that the "clean" points are: monitor_at_main,
and in the child after fork.  I'll see if I can adjust my patch with that in mind.

The first thing that hpcrun_prepare_measurement_subsystem does is use
an atomic variable for run once.  Lacking any documentation, that
misled me into thinking it was safe to call it other times.

mwkrentel:
@jmellorcrummey So, how does toolkit handle the case of a program creating 
a thread before main and launching gpu code from the tread?

For interrupt sources, you have to start the interrupts right away, or
else you don't get another chance to do so.
Is it different for gpu (or amd) sources?

jmellorcrummey:
> lSo, how does toolkit handle the case of a program creating a thread before main and launching gpu code from the tread?

monitor_init_thread calls prepare measurement subsystem if it is not a thread that should be ignored.

> |For interrupt sources, you have to start the interrupts right away, or
else you don't get another chance to do so.
Is it different for gpu (or amd) sources?

as described above.

blue42u:
[This repository has moved to GitLab, please continue this PR there.](https://gitlab.com/hpctoolkit/hpctoolkit/-/issues/631)

blue42u:
The previous link was incorrect, apologies. The corrected link is [here.](https://gitlab.com/hpctoolkit/hpctoolkit/-/merge_requests/631)

